PR #8, opened 2018-09-24
Style updates

angelw22:
Update and add logo,  add favicon, slight update to nav bar


yehjxraymond:
Few things I've noticed:
1. Max-width is not applied to to all elements/pages. Ie. "How it works" section and Registry page. 
2. Logo width is different for different institute. Like the bottom 4 logo takes up much less space than the top 4. Is this intentional?

yehjxraymond:
Also, seemed like main.scss is not used in `ViewerPageContainer.js`. We can remove that as well.

PR #10, opened 2018-09-26
Added print button and print watermark

yehjxraymond:


src/components/CertificateViewer.js
```diff
@@ -4,6 +4,7 @@ import CertificateVerifyBlock from "./CertificateVerifyBlock";
 import MultiCertificateRenderer from "./MultiCertificateRenderer";
 import templateRegistry from "./CertificateTemplates";
 import InvalidCertificateNotice from "./InvalidCertificateNotice";
+import styles from "./certificateviewer.scss";
```
yehjxraymond: I think you've forgotten to commit this file in.

angelw22: committed

PR #20, opened 2018-11-01
Share button 

yehjxraymond:


src/services/email/index.js
```diff
@@ -0,0 +1,21 @@
+import "isomorphic-fetch";
+
+export const emailApiUrl =
+  "https://92cdczopwg.execute-api.ap-southeast-1.amazonaws.com/dev";
```
rjchow: should probably move this to .env

src/config/index.js
```diff
@@ -1,7 +1,8 @@
 import { types as Web3Types } from "../services/web3/getWeb3";
 
 const config = {
-  DEFAULT_NETWORK: Web3Types.INFURA_ROPSTEN
+  DEFAULT_NETWORK: Web3Types.INFURA_ROPSTEN,
+  CAPTCHA_CLIENT_KEY: "6LfiL3EUAAAAAHrfLvl2KhRAcXpanNXDqu6M0CCS"
```
rjchow: this is public information but should probably put in .env to separate config

PR #24, opened 2018-11-02
Template/ngee ann polytechnic

yehjxraymond:


src/components/CertificateTemplates/NgeeAnnPolytechnic/NP-AA2018-OPTION/certificate.js
```diff
@@ -0,0 +1,170 @@
...
+    "August",
+    "September",
+    "October",
+    "November",
+    "December"
+  ];
+
+  const day = date.getDay();
+  let daySup = "";
+  switch (day % 10 === 1) {
```
rjchow: wat

rjchow: let's provide them with a date library to use, laurent suggests date-fns

src/components/CertificateTemplates/NgeeAnnPolytechnic/common/images.js
```diff
@@ -0,0 +1,38 @@
+export const IMG_LOGO_NP =
```
rjchow: why not have them as image files instead of b64 strings?

src/components/CertificateTemplates/NgeeAnnPolytechnic/common/index.js
```diff
@@ -0,0 +1,8 @@
+export * from "./images";
+export * from "./template";
+
+export const approvedAddresses = [
+  "0xa5d801265D29A6F1015a641BfC0e39Ee3dA2AC76",
```
rjchow: remove the unused ones and leave a comment indicating the network

src/components/CertificateTemplates/NgeeAnnPolytechnic/common/transcript.js
```diff
@@ -0,0 +1,409 @@
...
+  width: "100%",
+  height: "auto"
+};
+
+const thWidth60Left = {
+  width: "60%",
+  textAlign: "left"
+};
+
+const formatDate = dateString => {
```
rjchow: i think this is duplicated in cert

src/components/CertificateTemplates/NgeeAnnPolytechnic/common/images.js
yehjxraymond: Hmm allow them the access into the `static` folder? Might prefer not, think it might be easier to maintain if they only have one folder.

src/components/CertificateTemplates/NgeeAnnPolytechnic/common/index.js
yehjxraymond: Will need them to confirm which is which.

rjchow: resolve when they reply

PR #39, opened 2018-11-14
add new template for diploma plus certificate.

mmysgp:
NP: add a new template for diploma plus certificate.


yehjxraymond:
Hi Wu Yong,

We've merged the changes in from another branch, will be closing this. Could you allow maintainers to amend your branch next time: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

Also, a few suggestions: 
1. Prevent code duplication - There were still numerous code duplication found in the templates, we have fixed them. Prefer you break down templates and reuse existing components between templates.

2. Use of png images - Currently some of the images are using jpeg instead of png. They tend to have lower resolution, bigger and does not support alpha channel (transparency).

3. Scaling images - Some of the signatures are pretty big and pixelated. Making them smaller (fix width) may make it look neater

4. Use lodash.get - We noticed a couple of certificate.something.something[something]. If any of the element is not found, the site will throw a major error. Prefer you use something like get(certificate, "something.somthing.somthing", <default value>).

Issue #119, opened 2019-03-14
Preloading the template chunk for the certificate file


rjchow:
Templates are split into chunks for each institutions, currently these chunks are only loaded **after** blockchain verification step - it would be good to do this in parallel with the blockchain verification so that the user perceived loading time is shorter

PR #121, opened 2019-03-14
[#119] - Preloading the template chunk for the certificate file

sumitnitsurat:
- added the preloading of templates before the verification starts.

### DIFF
--- a/src/sagas/templates.test.js
+++ b/src/sagas/templates.test.js
@@ -1,0 +1,41 @@
+import { put, call } from "redux-saga/effects";
+import { get } from "lodash";
+import { certificateData } from "@govtechsg/open-certificate";
+import templates from "../components/CertificateTemplates";
+import { preloadTemplateChunk } from "./templates";
+import { types as templateActions } from "../reducers/templates";
+import { MakeCertUtil } from "./testutils";
+
+describe("preload template", () => {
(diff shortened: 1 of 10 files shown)


rjchow:
@sumitnitsurat please review and write tests when you get back 😜

src/sagas/templates.test.js
```diff
@@ -0,0 +1,36 @@
+import { put, call } from "redux-saga/effects";
+import { get } from "lodash";
+import { certificateData } from "@govtechsg/open-certificate";
+import templates from "../components/CertificateTemplates";
+import json from "../../tests_fixture/NP_Certs_FT_BMS.json";
```
rjchow: i think just use a minimal cert inline instead of using a full cert

rjchow (changes requested):

src/sagas/templates.test.js
sumitnitsurat: We have any function which returns the minimal cert or you want me to remove the data from this cert itself ?

rjchow: you can use this: https://github.com/OpenCerts/opencerts-website/blob/master/src/sagas/testutils.js

rjchow (approved):

PR #140, opened 2019-04-01
added a new template for ssgwsg-qual certs

ssgwsg-acn:


ssgwsg-acn:
we are resolving issues

PR #141, opened 2019-04-02
Fix/try another link

sumitnitsurat:
- Resetting the certificate state to initial state.


src/reducers/certificate.js
```diff
@@ -37,6 +37,7 @@ export const initialState = {
 
 // Actions
 export const types = {
+  RESET_CERTIFICATE: "RESET_CERTIFICATE",
```
yehjxraymond: Possible to use `UPDATE_CERTIFICATE` action with empty certificate instead? That way, no need for new action.

sumitnitsurat: Can not use "UPDATE_CERTIFICATE" directly as it calls verify certificate calls subsequently.

rjchow (changes requested):
add test

src/reducers/certificate.test.js
```diff
@@ -1,12 +1,23 @@
...
   getActiveTemplateTab,
-  getTemplates
+  getTemplates,
+  resetCertificateState
 } from "./certificate";
 
 describe("reducers", () => {
+  describe("Reset certificate", () => {
+    it("should reset the state to initial state", () => {
+      const action = {
```
rjchow: should use this instead
https://github.com/OpenCerts/opencerts-website/blob/e369fe2309bc03e4b4bdef083e3eed3a44aaa060/src/reducers/certificate.js#L303

sumitnitsurat: added test for action also.

rjchow: no need, just use the exported function directly instead of replicating the action generator here

rjchow (approved):

PR #168, opened 2019-04-16
update jest

rjchow:
add number-to-words lib for RP
update jest
fix broken tests surfaced by jest upgrade


src/sagas/testutils.js
```diff
@@ -43,6 +43,15 @@ export class MakeCertUtil {
...
   }
 
+  static makeFakeUUID() {
+    return "aaaaaaaa-aaaa-4aaa-aaaa-aaaaaaaaaaaa";
+  }
+
+  addTemplateTag(tag) {
+    this.cert.data.$template = `${MakeCertUtil.makeFakeUUID()}:string:${tag}`;
+    return this;
+  }
```
sumitnitsurat: addTemplateTag(field, tag) {
  this.cert.data[field]= `${MakeCertUtil.makeFakeUUID()}:string:${tag}`;
  return this;
 }

We can make it generic by passing the field to add the basic schema fields to test in future.

rjchow: did it this way so that you don't have to be concerned with the uuid when you're writing tests - in the future this method will probably calculate the hashes so that we can do tests without having to update all the hashes

PR #171, opened 2019-04-17
Moving handleObfuscation and certificate into MultiCertRenderer

yehjxraymond:
- Exposing `window.templates` as a function to return the array of templates available, helps to allow clients using frameless-viewer to determine the number of templates (and their names) available.
- Exposing `window.renderCertificate` as a function to render the certificate (in JSON) in the frameless-viewer
- Exposing `window.selectTemplateTab` as a function to switch between the different templates available in the frameless-viewer
- Connected the store to `MultiCertificateRenderer`, `certificate` and `handleObfuscation` is no longer needed to be passed in as props
- Updated all certificates using `MultiCertificateRenderer` to use the simpler component
- Added integration test to demo interacting with `frameless-viewer` with js injection


src/reducers/certificate.js
```diff
@@ -32,7 +32,10 @@ export const initialState = {
   verificationStatus: [],
 
   emailState: states.INITIAL,
-  emailError: null
+  emailError: null,
+
+  templates: null,
+  activeTemplateTab: 0
```
sumitnitsurat: Just a suggestion...Its better for state management if we keep UI state as local. Keep logical state in redux global state. Though our app is not that big so fine for now.

src/components/FramelessViewerPageContainer.js
```diff
@@ -24,6 +27,12 @@ class FramelessViewerContainer extends Component {
     this.handleTextFieldChange = this.handleTextFieldChange.bind(this);
   }
 
+  componentDidMount() {
+    window.templates = () => this.props.templates;
+    window.renderCertificate = this.handleCertificateChange;
```
rjchow: i think we shouldn't unnecessarily pollute the global namespace, can we put it under windows.opencerts.xx?

src/reducers/certificate.js
rjchow: i would ordinarily agree with this, but the alternative is to propagate the handler for setting tab index through the whole hierarchy and i think that's a worse idea, open to discussion on this

PR #176, opened 2019-04-22
Fix/seo optimization

thefiend:
Added meta information, added alt text to images and hyperlink to external webpages


rjchow:
leaving this to raymond

thefiend:
Just to check, is Meta.js used anywhere in the project or is it not required? Can't seem to find any usage for that.

yehjxraymond:
Yep you are right, was previously called in every page but think it was removed later. Can remove that for now

pages/_document.js
```diff
@@ -20,6 +20,27 @@ export default class MyDocument extends Document {
           />
           <meta name="viewport" content="width=device-width, initial-scale=1" />
           <meta charSet="utf-8" />
+          <meta
+            name="description"
+            content="Whether you're a student or an employer, OpenCerts lets you verify the certificates you have of anyone from any institution. All in one place."
```
yehjxraymond: Description should be different for different pages. Might want to consider having this meta in the individual pages instead.

pages/_document.js
```diff
@@ -20,6 +20,27 @@ export default class MyDocument extends Document {
...
+          />
+          <meta
+            name="keywords"
+            content="OpenCerts, GovTech, Check, Verify, Certificate, Blockchain"
```
yehjxraymond: Keywords are hardly used in modern SEO, but it won't hurt.

src/components/CertificateDropZone/Views/DefaultView.js
```diff
@@ -11,7 +11,10 @@ const View = ({ hover, accept }) => (
   >
     <div className={css["image-container"]}>
       <i>
-        <img src="/static/images/dropzone/dropzone_illustration.svg" />
+        <img
+          alt="Dropzone"
```
yehjxraymond:
```suggestion
          alt=".opencert Dropzone"
```

Might want to consider above, or a variation. Reason is to help people figure out "how to open .opencert".

pages/_document.js
yehjxraymond: Overall I'm thinking of a `Meta` component where you can pass in these as props. So each page will have something like:
```js
<Meta
  description="Some description"
  keywords="A, B, C"
/>
```

As I'm writing this, it seemed like there is a package doing exactly this, you might want to check out: https://www.npmjs.com/package/next-seo

pages/faq.js
```diff
@@ -1,18 +1,33 @@
+import React from "react";
 import { connect } from "react-redux";
-import Head from "next/head";
+import NextSeo from "next-seo";
 import NavigationBar from "../src/components/Layout/NavigationBar";
 import FooterBar from "../src/components/Layout/FooterBar";
 import FaqContent from "../src/components/FAQ/FaqContent";
 
+import { SEO } from "../src/config";
```
yehjxraymond: Consider naming it SEO_DEFAULT or something to tell its default.

pages/faq.js
```diff
@@ -1,18 +1,33 @@
...
+
+export const FAQ_SEO = {
```
yehjxraymond: 1. Perhaps can rename to just SEO or META, easier to copy across file lol.

2. Consider using lodash deep merge, then no need to recurs inside object.
https://lodash.com/docs/4.17.11#merge

```
const _ = require("lodash")
_.merge({a: {b:2, c: 4}}, {a: {b:7}})
> { a: { b: 7, c: 4 } }
```

yehjxraymond:
```
merge(SEO_DEFAULT, {
  title: ...
  description: ...,
  openGraph: {...}
})
```

pages/faq.js
```diff
@@ -1,18 +1,33 @@
...
+    url: "https://opencerts.io/faq"
+  }
+};
+
 const FaqPage = () => (
-  <div>
-    <Head>
-      <title>OpenCerts - FAQ</title>
-    </Head>
+  <React.Fragment>
```
yehjxraymond: Found out there is a shorthand from RJ recently, you can try <> and </> to represent React fragment

pages/privacy.js
```diff
@@ -1,18 +1,33 @@
...
 import { connect } from "react-redux";
-import Head from "next/head";
+import NextSeo from "next-seo";
 import NavigationBar from "../src/components/Layout/NavigationBar";
 import FooterBar from "../src/components/Layout/FooterBar";
 import PrivacyContent from "../src/components/Privacy";
 
+import { SEO } from "../src/config";
+
+const _ = require("lodash");
```
yehjxraymond:
```suggestion
import { merge } from "lodash"
```

pages/faq.js
```diff
@@ -1,18 +1,32 @@
...
-import Head from "next/head";
+import { merge } from "lodash";
+import NextSeo from "next-seo";
 import NavigationBar from "../src/components/Layout/NavigationBar";
 import FooterBar from "../src/components/Layout/FooterBar";
 import FaqContent from "../src/components/FAQ/FaqContent";
 
+import { SEO } from "../src/config";
+
+merge(SEO, {
```
yehjxraymond: Didn't realise this is an unpure function. This way you potentially alter the state of SEO for other pages. The fix can be:

```
const SEO = merge({
  <your overwrite values here>
}, DEFAULT_SEO)
```

Consider use DEFAULT_SEO for your config, so it's obvious it's a default, and then assign SEO to this var instead.

yehjxraymond: Switching the order of the args around will not alter the second obj, in case you are wondering what's the diff. But test it out yourself though.

pages/index.js
```diff
@@ -1,18 +1,19 @@
...
 
+import { DEFAULT_SEO } from "../src/config";
+
 const VerifierPage = () => (
-  <div>
-    <Head>
-      <title>OpenCerts</title>
-    </Head>
+  <>
+    <NextSeo config={DEFAULT_SEO} />
```
sumitnitsurat: Can you check if we can move in higher component which is common for every view ? I think we can move it in _app.js.

yehjxraymond: He needs to be able to customise this based on different page. Any suggestions?

sumitnitsurat: create a json of SEO configuration based on different pages routes. keep only one merge function and pass value based on the routes you are in. it will be really easy to maintain.

yehjxraymond: Hmm that way when creating a single page, I've one additional file to edit?

sumitnitsurat: No..by default a default SEO configuration will be applied on the newly created page..and if you want to change you just need to add the config for that page ..Thats it..you dont need to include <NextSeo> component and write merge for that configuration again.  The best thing if later on we want to change the merge function or add props in <NextSeo> component we wont need to take care of every place we have included it.

yehjxraymond: Oh I see, so if you like the default, then don't need to pass in the extra params? I like.

pages/_app.js
```diff
@@ -23,7 +71,10 @@ class MyApp extends App {
     return (
       <Container>
         <Provider store={store}>
-          <Component {...pageProps} />
+          <>
```
sumitnitsurat: Dont need this open and close brackets.

pages/_app.js
```diff
@@ -1,13 +1,61 @@
...
+import NextSeo from "next-seo";
 import withGA from "next-ga";
+import { merge } from "lodash/fp";
 import React from "react";
 import { Provider } from "react-redux";
 import initStore from "../src/store";
-import { GA_ID } from "../src/config";
+import { GA_ID, DEFAULT_SEO } from "../src/config";
+
+const ALL_SEO_CONFIG = {
```
sumitnitsurat: move this to config only with default SEO. name const as SEO and make a key "DEFAULT" and then rest of the routes.

PR #198, opened 2019-04-24
Separated registry networks to Main and Ropsten

thefiend:
Separated the networks in registry to respective networks and added checking for different network types


rjchow:
was talking about this with raymond, I think perhaps we should combine the data for our viewable registry page and this - we do need to get the contact details for the institutes though.

Alternatively we can just create a card that shows only the details that exist and slowly fill in the details

thefiend:
> was talking about this with raymond, I think perhaps we should combine the data for our viewable registry page and this - we do need to get the contact details for the institutes though.
> 
> Alternatively we can just create a card that shows only the details that exist and slowly fill in the details

Yea, talked to Raymond about it before. We probably can check if the URL for the partner's OpenCert exists in the registry, if it does then show on the registry page.

PR #199, opened 2019-04-24
Added admin panel and documentation to home page footer

thefiend:


rjchow:
can I ask what's the rationale for this?
don't think it's necessary to expose it to cert users as the admin side is quite separate from the viewer side

sumitnitsurat:
> can I ask what's the rationale for this?
> don't think it's necessary to expose it to cert users as the admin side is quite separate from the viewer side
Its just my opinion, I think it would be nice to have a reference for all the links in our website rather than going and accessing the docs from Github. If we will give a visibility to the "OpenCerts" as a product. Though if we dont want to expose admin panel and docs to everyone then we can remove it. your call.

rjchow:
just put the docs link in the faq if its not already there, and no need for the admin page link

PR #204, opened 2019-04-25
Fixed encoding for FileReader

yehjxraymond:


yehjxraymond:
[example.0.txt](https://github.com/OpenCerts/opencerts-website/files/3115016/example.0.txt)
[sample.txt](https://github.com/OpenCerts/opencerts-website/files/3115017/sample.txt)


Can try with these two certs.

PR #209, opened 2019-04-25
 Added common fonts and styling to templates (Basic SOA, QUAL and TRANS)

ssgwsg-acn:
---
name: Certificate Template Addition
about: This is the workflow for requesting new certificate templates to be added to
  the OpenCerts repository
title: "[New Template]"
labels: new template
assignees: ''

---

# Pull Request Guidelines for Adding Certificate Templates
This document is a work in progress but here are some basic checks. As these are only basic guidelines, meeting the below doesn't indicate there will be no issues with your pull request.

### Certificate Template 
- [ ] No more than 5 templates or 25 added/modified files in the pull request
- [ ] Integration test for each template that checks that the correct rendering is done given a sample certificate
- [ ] Sample certificate file included for each template, located alongside the integration test for each template
- [ ] Sample certificates must obviously be a sample certificate
  - [ ] Obviously fictitious name
  - [ ] Obviously sample signatory images
- [ ] No fixed-size raster images as part of certificate layout
- [ ] Mobile responsive design
- [ ] Date parsing should be localised to template author's timezone
- [ ] Webpack chunking code is correct
  - [ ] Has chunking code
  - [ ] Same chunking code as the other certificates belonging to that institute
- [ ] Certificate Store Addresses have been updated
  - [ ] Template Whitelist
  - [ ] Registry
- [ ] Template should not be using resources(images etc.) on the website outside of their own folder (e.g institute logo shouldn't be used from /static because there's no guarantee it will not change)

### Pre-merge checks

- [ ] Ensure that your code has been [rebased](https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request) on top of latest OpenCerts master
- [ ] Linter issues resolved (Run `npm run lint:fix` to see issues)
- [ ] `npm run test` passes
- [ ] `npm run test:integration` passes
- [ ] [Travis Build passes](https://docs.travis-ci.com/user/for-beginners/)


src/components/CertificateTemplates/tlds/sg/gov/ssg-wsg/common/fonts.js
```diff
@@ -0,0 +1,30 @@
+const fonts = () => (
+  <div>
+    <head>
+      <link
+        href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,600"
+        rel="stylesheet"
+      />
+    </head>
+    <style jsx global>{`
```
rjchow: i believe this global keyword will affect things outside your template

yehjxraymond: You can use a stylesheet for this. Please see how other schools use extra stylesheets, can check out NP

ssgwsg-acn: rjchow are you sure about this because this fonts function only using by our templates? can you share us the details how in opencerts templates we can use custom fonts? because this looks like common item which everyone should know how they can implement custom fonts in Opencerts?.

It will be helpful if you have this documented as part of your docs.opencerts.io.  So everyone will follow same standard instead of changing approach again.

rjchow: https://github.com/zeit/next.js/wiki/Global-styles-and-layouts
As you can see it modifies the <head> element on the top level page which is the entire viewer page

The last person who requested custom fonts was told that we do not support custom fonts, but we will be investigating to see how we can support it. You're the first entity to be allowed to use custom fonts so there is no instructions or standards yet. If you don't feel like doing a little bit of work to figure out how to do it properly, feel free to not use custom fonts as is the official stand right now.

In any case I don't see why you're having so much trouble with this, all you need to do is insert the <link> element somewhere in the template instead of using <style jsx global>, it will apply just fine without being in <head>

yehjxraymond:
Noticed the sonar build failure, looking into it

ssgwsg-acn:
We have added the custom font styling and its working fine.
Also, We have refactored the methods as per sonar requirement.

rjchow:
thanks for making the changes, appreciate it

PR #213, opened 2019-04-26
Fix/verified button

sumitnitsurat:
- Verified green button added on mobile. 
- Lint fix changes are also there.


yehjxraymond (approved):
Looks good to me (=

PR #225, opened 2019-04-30
Fix/distorted icons

sumitnitsurat:


src/components/CertificateVerifyBlock/certificateVerifyBlock.scss
```diff
@@ -74,4 +74,7 @@
   .d-flex {
     display: block!important;
   }
+  .simple-verify-block {
+    min-width: 250px;
```
yehjxraymond: Instead of min width, can we use the bootstrap class to go for full width on small device?

src/components/CertificateViewer.js
```diff
@@ -33,40 +33,45 @@ const renderHeaderBlock = props => {
...
-              title="Download"
+              id="btn-print"
+              className={styles["print-btn"]}
+              onClick={() => window.print()}
             >
-              <i
-                className="fas fa-file-download"
-                style={{ fontSize: "1.5rem" }}
-              />
+              {images.print()}
```
yehjxraymond: Shall we use FontAwesome and then no need to be bothered about the padding issue on multiple devices?

src/components/CertificateVerifyBlock/certificateVerifyBlock.scss
```diff
@@ -69,12 +71,15 @@
...
+    margin: 0px 15px 15px 0;
   }
   .d-flex {
     display: block!important;
   }
-  .simple-verify-block {
-    min-width: 250px;
+}
+
+@media only screen and (min-width: 550px) and (max-width: 768px) {
```
yehjxraymond: Prefer to use bootstrap classes instead of media queries? My concern is that the breakpoint might not be aligned with those for bootstrap

sumitnitsurat:
@yehjxraymond Its good to merge in master ?

yehjxraymond:
Yes please (:

PR #266, opened 2019-05-13
Update FaqContent.json

rjchow:
add developed by govtech

need to enable html content

### DIFF
--- a/src/components/FAQ/FaqContent.js
+++ b/src/components/FAQ/FaqContent.js
@@ -1,5 +1,4 @@
 import PropTypes from "prop-types";
-import Link from "next/link";
 import styles from "./faq.scss";
 import content from "./FaqContent.json";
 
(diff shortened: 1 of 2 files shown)


src/components/FAQ/FaqContent.js
```diff
@@ -9,21 +8,14 @@ const FaqHeader = () => (
...
       <div className={styles.answer}>
-        {answer}{" "}
-        {url ? (
-          <Link href={url}>
-            <a>{url}</a>
-          </Link>
-        ) : (
-          ""
-        )}
+        <span dangerouslySetInnerHTML={{ __html: answer }} />
```
sumitnitsurat: What about sanitising the content before setting it?

rjchow: we write the content

src/components/FAQ/FaqContent.json
```diff
@@ -28,16 +32,15 @@
       },
       {
         "question": "Why can't I print the certificate?",
-        "answer": "Printing the certificate discards all the advanced cryptographic protections we have built into OpenCerts, hence printing is not recommended. "
+        "answer": "Printing the certificate discards all the advanced cryptographic protections we have built into OpenCerts, hence printed certificates are not to be considered authentic."
```
sumitnitsurat: "hence printed certificates are not considered authentic".

rjchow: hmmmmmm........ very minor semantic difference but it was deliberate;
"not considered authentic" -> its not authentic for sure
"not to be considered authentic" -> it may be authentic but you don't know if it is and you cannot know, because there's no cryptographic proof

most people wont notice the difference though 😂

PR #275, opened 2019-05-15
Browser back btn

sumitnitsurat:
- For this we need web3 unit testing. Will create a separate issue for that.


rjchow:
i wrote a quick test here that reproduces the issue, https://github.com/OpenCerts/opencerts-website/commit/fc1912e099376affd69bc853fd3fdce5dcb66bb4

use it as reference to write a regression test for this but don't use Ngee Ann's file for this because we don't control their cert store - use our testnet store to issue a test certificate, i'm not sure perhaps the example cert or the default cert are already issued? get the ledger nano from me to do the issuance

rjchow:
`npx testcafe chrome ./src/services/web3/integration.spec.js`

also write a npm command that runs integration tests in the current working directory

src/components/CertificateVerifyBlock/integration.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    async (prev, curr) => t.expect(component.textContent).contains(curr),
+    Promise.resolve()
+  );
+
+test("DEMO certificate is rendered correctly", async t => {
+  // Uploads certificate via dropzone
+  await t.setFilesToUpload("input[type=file]", [Certificate]);
+
+  // Certificate tabs rendered
+  //   await t.expect(TemplateTabList.textContent).contains("Certificate");
```
yehjxraymond: Why use comments here?

src/components/CertificateVerifyBlock/integration.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    "MIND QUEST FOR EXCELLENCE",
+    "TRF",
+    "2",
+    "PX",
+    "C+"
+  ]);
+
+  await ClientFunction(() => window.history.back())();
+  await t.setFilesToUpload("input[type=file]", [Certificate]);
+  //   await t.expect(TemplateTabList.textContent).contains("Certificate");
```
yehjxraymond: Remove comment pls

yehjxraymond (approved):

PR #279, opened 2019-05-16
Added message passing capabilities for frameless-viewer for frameless-viewer

yehjxraymond:


yehjxraymond:
@rjchow suggested to test out using a lib for the message passing:
https://github.com/Aaronius/penpal
https://github.com/dollarshaveclub/postmate

src/components/FramelessViewer/FramelessViewerPageContainer.js
```diff
@@ -6,33 +6,65 @@ import {
...
+  getActiveTemplateTab,
+  selectTemplateTab as selectTemplateTabAction
 } from "../../reducers/certificate";
 import FramelessCertificateViewer from "./FramelessCertificateViewer";
 import { getLogger } from "../../utils/logger";
 
 const { trace } = getLogger("components:FramelessViewerPageContainer");
 
+const inIframe = () => window.location !== window.parent.location;
+const flatten = o => JSON.parse(JSON.stringify(o));
```
rjchow: if its only necessary for the parent to know which templates exist, use lodash.pick to grab just the id and label - that way we don't accidentally expose information in the future when template objects change

also rename this method so that it makes sense

src/components/FramelessViewer/FramelessViewerPageContainer.js
```diff
@@ -6,33 +6,65 @@ import {
...
-      selectTemplateTab: this.props.selectTemplateTab
+      getTemplates,
+      renderCertificate,
+      selectTemplateTab
     };
+
+    if (inIframe()) {
+      const parentFrameConnection = connectToParent({
+        methods: {
+          renderCertificate,
```
rjchow: should probably do some defensive sanitization on all the parameters for the methods below since they execute

not that i can think of something they couldn't already do since its all clientside js, but seems like good practice?

rjchow: also document the methods (what they do and the data structure for the parameters they take/return)

src/components/FramelessViewer/iframe.html
```diff
@@ -0,0 +1,580 @@
...
+    const iframe = document.getElementById("frameless-iframe");
+    var iframeMethods;
+
+    selectTemplateTab = function(i) {
+      window.connection.promise.then(frame => frame.selectTemplateTab(i));
+    };
+
+    updateHeight = function(h) {
+      // Adding 60 to account for extra height on firefox
+      iframe.height = h + 60;
```
rjchow: don't know if it matters since this is a demo implementation - but this line is causing the page height to continue increasing if you click on render repeatedly

src/components/FramelessViewer/iframe.html
```diff
@@ -0,0 +1,580 @@
...
+    };
+
+    updateTemplates = function(templates) {
+      if (!templates) return;
+      const templateSelector = document.getElementById("template-selectors");
+      templateSelector.innerHTML = "";
+      templates.forEach((t, i) => {
```
rjchow: probably want to be clearer with the variable names since this is for people to reference

src/components/FramelessViewer/FramelessViewerPageContainer.js
```diff
@@ -6,33 +6,65 @@ import {
...
+        if (parent.updateTemplates)
+          parent.updateTemplates(flatten(this.props.templates));
+      });
+    }
   }
 
   componentDidMount() {
+    const { selectTemplateTab } = this.props;
+    const getTemplates = () => flatten(this.props.templates);
+    const renderCertificate = this.handleCertificateChange;
```
rjchow: weren't we going to do verification against the documentstore? I think this doesn't - i tried modifying the merkle root of the input cert and it still passes

src/components/FramelessViewer/FramelessViewerPageContainer.js
yehjxraymond: Renamed to `formatTemplate`

src/components/FramelessViewer/FramelessViewerPageContainer.js
yehjxraymond: Cant think of the types of sanitization to perform right now. Thinking that it can be deferred as they do not perform an attack on our site, but rather on their implementation.

Have added the documents... felt like an essay

src/components/FramelessViewer/iframe.html
yehjxraymond: Removed, last check against firefox didnt have this issue.

src/components/FramelessViewer/FramelessViewerPageContainer.js
yehjxraymond: Not doing verification, verification should be done on parent side. Not going to babysit them through.

src/components/FramelessViewer/iframe.html
yehjxraymond: Fixed

PR #286, opened 2019-05-21
Soften the warning for unregistered institutes

rjchow:
Removed the warning dialog on unrecognised issuer check
Added a warning status icon for unrecognised issuer

Outstanding work in separate story:
- refactor the verifying stuff so that the status checks are not duplicated everywhere
- move Router.push to somewhere more suitable if possible


src/components/CertificateVerifyBlock/DetailedCertificateVerifyBlock.js
```diff
@@ -2,54 +2,66 @@ import PropTypes from "prop-types";
...
+  return isVerified ? renderSuccess(type) : renderFailure(type);
+};
+
 const CHECKS = {
   HASH: {
     id: "hashStatus",
     success: "Certificate has not been tampered with",
-    failure: "Certificate has been tampered with"
+    failure: "Certificate has been tampered with",
+    failureStatusIcon: FailureIcon
```
yehjxraymond: Why not just have the `failureStatusIcon` as icon? Then pass the entire `check` object into `CheckStatusRow`.

yehjxraymond: Also Instead of `FailureIcon`, was thinking if you want to just pass in the object `<i className="fas fa-times-circle text-danger mr-2" />`?

rjchow: i purposely extracted the failure flow out so that it's easy to read what's going on, and easier to change in the future

yehjxraymond: Hmm it was a little hard to read the code when i've to jump around. Looking at it again, was thinking if the interface of the `CheckStatusRow` can just be `check`. Then to render success or failure, it will all be in `CheckStatusRow`.

rjchow: mixing the logic and the html is harder to read for me, the icon stuff will get better after installing the fontawesome js libs that i did in another PR and do a bit more refactoring - just rushing this out because barry wants it

yehjxraymond: Haha okay

PR #293, opened 2019-05-22
Try opencerts demo

sumitnitsurat:
This issue was created because we would like to create a demo certificate for new users to experience the features of OpenCerts. This included a draggable demo certificate on the home page, which would be resized to a demo button for mobile view.

Once clicked, the new user would be shown 3 tabs - certificate, transcript, and media. They would be able to toggle with the tabs and play with the features of OpenCerts, such as data obfuscation and playing media.


rjchow:
closing to test something

PR #316, opened 2019-06-06
added ga + fixed draggable

samuellye98:


rjchow:
noticed #329 but this PR is good to be merged

PR #321, opened 2019-06-07
feat(Registry): new registry format

samuellye98:
This issue was created because we did not want duplicate registries - for the verification of issuer identity and the registry page for the MAIN net. This would help increase efficiency and reduce duplication of code. 

We changed registry.json to RegistryInfo.json such that the Registry page and sagas/certificate.js now access the same registry information. RegistryInfo.json no longer separated MAIN and ROPSTEN networks. Instead, it differentiates which information to show based on the property "displayCard". If true, it belongs to MAIN, ROPSTEN otherwise. 

We also created a test for the registry page - integration-registry.spec.js  - to ensure the content shown there matches RegistryInfo.json.


src/components/RegistryPageContainer.js
```diff
@@ -1,34 +1,31 @@
...
-    address: "0xa5d801265D29A6F1015a641BfC0e39Ee3dA2AC76",
-    website: "https://www.np.edu.sg",
-    email: "asknp@np.edu.sg",
-    phone: "+65 6466 6555",
-    key: "np",
-    logo: "/static/images/NP_logo.svg"
-  }
-];
+const members = [];
+Object.keys(registry.issuers).map(issuer =>
```
rjchow: lodash filter will do this much tidier
```
filter(registry.issuers, "displayCard")
```

src/services/issuers/index.js
```diff
@@ -1,4 +1,5 @@
-const defaultUrl = "/static/registry.json";
+const defaultUrl = "/static/RegistryInfo.json";
+// const defaultUrl = "/static/registry.json";
```
rjchow: any reason why this is left here? we usually just delete stuff that isn't used anymore

static/RegistryInfo.json
```diff
@@ -0,0 +1,112 @@
+{
```
rjchow: reckon we can just continue using registry.json instead of making a new file

src/components/RegistryPageContainer.js
```diff
@@ -1,34 +1,31 @@
...
+          src={m.logo}
+          id={css[m.key]}
+          style={{
+            maxHeight: "80px",
+            height: "100%",
+            width: "auto"
+          }}
+        />
         <h4 className={css["partner-name"]}>{m.name}</h4>
         {m.address ? (
```
rjchow: the registry addresses aren't showing up in the card, also reckon we should probably write a test for the card component?

rjchow:
image

bit of overflow on mobile but I don't really know of a good solution for this, anyone has any ideas on a reasonable way to handle this?

src/components/RegistryPageContainer.js
samuellye98: Added registry test

src/components/RegistryPageContainer.js
```diff
@@ -13,8 +13,9 @@ const renderMembers = () =>
   finalmembers.map((m, i) => (
     <div
       key={i}
+      id={m.key}
```
rjchow: div#institute is a bit too general of a id, probably want to use a more selective id name so that the namespace doesn't get cluttered, especially div#govtech

Nebulis: Feel free to use css modules to avoid such issues. otherwise it's important to be careful on named use (especially with id as they are supposed to be uniq, and when it's not the case weirds bugs may appear :) )

Nebulis:
https://css-tricks.com/almanac/properties/o/overflow-wrap/

src/components/RegistryPageContainer.js
samuellye98: changed to id={m.key}-registry

static/RegistryInfo.json
rjchow: ^ why's RegistryInfo.json is preferred over the existing registry.json?

samuellye98: o it was to differentiate between the 2 files at first but i'll change it back now

rjchow: not changing the filename preserves git history and helps with continuity - that way in the future if someone looks at it and wonders how the registry entries came to be, they can look at the git history

rjchow:
merge whenever :)

PR #323, opened 2019-06-07
fix(iframe): Remove iframe header

samuellye98:
This issue was created because we would like to remove the navbar for viewing certificates using the frameless viewer. This is because the iframe already has options to toggle through the certificate and transcript tab, thus the navbar was just a redundant repetition.

We removed the iframe header by editing certificateViewer.scss and creating a parent div for the frameless header - where we set the child div of that parent div to display: none


rjchow:
what issue does this resolve?

samuellye98:
> what issue does this resolve?
 
Raymond wanted me to remove the header for the framless viewer
https://github.com/orgs/OpenCerts/projects/1#card-22559506

rjchow:
raymond wants you to remove the header becaaaaaaaaaaause? i mean in this case its obvious but in the general case it's a bit hard for me to review a PR that i have no idea the intent of

will touch on this tomorrow in the tech sharing discussion

yehjxraymond (approved):

PR #325, opened 2019-06-08
feat(page): Added collaborate page and linked in navigation bar

thefiend:
#324

PR #328, opened 2019-06-10
Feat/togglenav

samuellye98:
This issue was created because we realized there was no nav bar for mobile view. However, after discussion we realized that the user may be confused with the navbar in the mobile view and the navbar for viewing certificates. They would think that the navbar for viewing certificates would be the same as that of the home page and not open it even though it contains the different tabs for their certificates. Therefore, this issue was closed.


samuellye98:
Do not merge until #325 is merged

sumitnitsurat:
check the /renderer route page also...the current navigation bar is getting overlapped with the existing one.

PR #331, opened 2019-06-11
fix(Error): priority of displaying error messages

samuellye98:
This issue was created because we would like to prioritize the display of error messages so as to make the error messages shown more helpful to the user. We also add an additional error message if the user inputs an invalid store address.

The first method is by implementing a stack and using the error message with the highest priority. 

Use Redux to see the state of the error messages


src/sagas/certificate.js
```diff
@@ -77,6 +79,37 @@ export function* loadCertificateContracts({ payload }) {
   }
 }
 
+export function* verifyCertificateStore({ certificate }) {
```
rjchow: haha what's going on in here?

samuellye98: I created a new state (VERIFYING_CERTIFICATE_STORE_SUCCESS) to verify whether the store address exists so this function changes the state of the application based on this

src/sagas/certificate.js
```diff
@@ -77,6 +80,41 @@ export function* loadCertificateContracts({ payload }) {
...
+  try {
+    const data = certificateData(certificate);
+    const contractStoreAddresses = get(data, "issuers", []).map(
+      issuer => issuer.certificateStore
+    );
+
+    // Checks if issuing institution has a valid smart contract with OpenCerts
+    const bytecode =
+      "0x7135575eac76f1817c27b06c452bdc2b7e1b13240797415684e227def063a127";
+    const prom = yield web3.eth.getCode(contractStoreAddresses[0]);
```
rjchow: need to work on your variable names :p

bytecode, prom, and keccak - none of them are informative regarding what's happening over here

some things to think about:
what's the data that this label (variable name) refers to?
what's the purpose of this variable?
if someone else is reading this code, how can i help him understand what's going on here?

so for example a translation of this method into english, this verifyCertificateStore method:

get a web3provider so that i can interact with web3 stuffs
retrieve certificate contents from the opencert document
get all of the issuer's certificate store addresses
retrieve the bytecode of the smart contract on-chain at .. the first contract store address [1]
calculate the keccak256 hash of the bytecode that we retrieved
check if the keccak256 of the bytecode of the smart contract on-chain is the same as what we expect
throw an error if it doesn't[2]

check if first contractStoreAddress is a valid ethereum address [3]
put result into redux state
return true

throw error if it's not an ethereum address

----------
Technical issues
[1] why only the first issuer address? shouldn't it be checking all of the issuers?
[2] this is more of a smart contract isn't what we expect rather than smart contract doesn't exist
[3] does it make sense to check if its a smart contract address **after** retrieving the bytecode?

src/sagas/certificate.js
```diff
@@ -360,6 +401,15 @@ export function* analyticsRevocationFail({ certificate }) {
   });
 }
 
+export function* analyticsStoreFail({ certificate }) {
+  yield analyticsEvent(window, {
+    category: "CERTIFICATE_ERROR",
+    action: get(certificate, "issuers[0].certificateStore"),
```
rjchow: shouldn't be just referencing only the first issuer here either

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -1,14 +1,52 @@
...
 const View = ({
   resetData,
   issuerIdentityStatus,
   hashStatus,
   issuedStatus,
-  notRevokedStatus
+  notRevokedStatus,
+  storeStatus
 }) => {
+  const errorMessages = [
```
rjchow: separate this into another PR from the smart contract verification, this one's got its own complexities that we'll discuss separately

package.json
```diff
@@ -53,6 +53,7 @@
     "redux-devtools-extension": "^2.13.7",
     "redux-saga": "^0.16.0",
     "web3": "^1.0.0-beta.52",
+    "web3-eth": "^1.0.0-beta.55",
```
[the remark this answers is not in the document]
samuellye98: will remove - was looking at it but didn't need to use it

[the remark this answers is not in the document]
rjchow: for the scope of our interactions with web3 it might make more sense to use this smaller package but we can revisit that in a separate PR

src/sagas/certificate.js
samuellye98: [1] Changed to --> const contractStoreAddress = get(data, "issuers[0].certificateStore");
This reads the certificateStore property from the .opencert/.json file and assigns it to the constant. There should only be 1 store address listed in the .opencert file.

[2] Changed error message to "Invalid smart contract"

[3] Changed the order so it validates whether it is a valid address before retrieving the bytecode

rjchow: [1] unfortunately a cert can have any number of issuers :( it's a bit of a pain to work with that i know
[2] ok
[3] ok

src/sagas/certificate.js
```diff
@@ -77,6 +80,39 @@ export function* loadCertificateContracts({ payload }) {
...
+
+    // Checks if certificate has a valid certificate store address
+    if (!web3.utils.isAddress(contractStoreAddress)) {
+      throw new Error("Invalid certificate store address");
+    }
+
+    // Checks if issuing institution has a valid smart contract with OpenCerts
+    const supportedKeccakHash =
+      "0x7135575eac76f1817c27b06c452bdc2b7e1b13240797415684e227def063a127";
+    const certificateByteCode = yield web3.eth.getCode(contractStoreAddress);
```
rjchow: onChainByteCode

src/sagas/certificate.js
```diff
@@ -77,6 +80,39 @@ export function* loadCertificateContracts({ payload }) {
...
+    const data = certificateData(certificate);
+    const contractStoreAddress = get(data, "issuers[0].certificateStore");
+
+    // Checks if certificate has a valid certificate store address
+    if (!web3.utils.isAddress(contractStoreAddress)) {
+      throw new Error("Invalid certificate store address");
+    }
+
+    // Checks if issuing institution has a valid smart contract with OpenCerts
+    const supportedKeccakHash =
```
rjchow: make this an array and name it supportedContractHashes
(we're going to have to support multiple smart contract versions very soon, so just to save the rework)

src/sagas/certificate.js
```diff
@@ -77,6 +80,39 @@ export function* loadCertificateContracts({ payload }) {
...
+      "0x7135575eac76f1817c27b06c452bdc2b7e1b13240797415684e227def063a127";
+    const certificateByteCode = yield web3.eth.getCode(contractStoreAddress);
+    const certificateKeccakHash = web3.utils.keccak256(certificateByteCode);
```
rjchow: hashOfOnChainByteCode

src/sagas/certificate.js
```diff
@@ -77,6 +135,74 @@ export function* loadCertificateContracts({ payload }) {
   }
 }
 
+export function* isValidENSDomain(storeAddress) {
+  trace(`Checking if ${storeAddress} is a valid ENS Domain`);
+  const web3 = yield getSelectedWeb3();
+  const ensToAddress = yield web3.eth.ens.getAddress(storeAddress);
+  // console.log("IMPORTANT", ensToAddress);
```
rjchow: ^

src/sagas/certificate.js
```diff
@@ -182,9 +188,9 @@ export function* verifyCertificateStore({ certificate }) {
 
     // Checks if issuing institution has a valid smart contract with OpenCerts
     yield combinedStoreAddresses.map(address => isValidSmartContract(address));
-
     yield put(verifyingCertificateStoreSuccess());
-    return true;
+    return combinedStoreAddresses;
+    // return true;
```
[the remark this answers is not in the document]
samuellye98: it returns all the valid store addresses

src/sagas/certificate.js
```diff
@@ -77,6 +135,74 @@ export function* loadCertificateContracts({ payload }) {
...
+    );
+
+    // Concat the 2 arrays
+    const combinedStoreAddresses = compact(
+      ethereumAddressIssuers.concat(resolvedEnsNames)
+    );
+    // console.log("COMBINED", combinedStoreAddresses);
+
+    // Checks if issuing institution has a valid smart contract with OpenCerts
+    yield combinedStoreAddresses.map(address => isValidSmartContract(address));
```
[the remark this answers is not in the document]
samuellye98: Throws an error if there is a wrong smart contract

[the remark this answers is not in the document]
rjchow: yea but i dont know if isValidSmartContract actually executes if there's no call(), I have a feeling it doesn't - can you verify that it actually runs?

src/sagas/certificate.js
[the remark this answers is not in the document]
rjchow: // return true 

line 193 doesn't return anything

src/sagas/certificate.js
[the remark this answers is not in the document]
samuellye98: yup when I took it out and tested with a valid address but wrong smart contract, it did not display "Invalid Store". Only when I put it back did it show "Invalid Store"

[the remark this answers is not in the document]
rjchow: you're right it does run, odd I thought you would need a call()

PR #341, opened 2019-06-19
feat: Priority of error messages

samuellye98:
This PR was created because the current error message system displayed all the error messages that the redux state showed. At times, an invalid certificate would show all 4 error messages when the error was only due to failure in 1 state, which caused the rest of the states to fail as well. This would be of no help to the user as they would not understand which error message to tackle. 

Therefore, the first variation of the fix would be to create a stack and pop the error message off according to priority. This occurs in UnverifiedView.js

### DIFF
--- a/src/components/CertificateDropZone/Views/UnverifiedView.js
+++ b/src/components/CertificateDropZone/Views/UnverifiedView.js
@@ -1,5 +1,6 @@
 import PropTypes from "prop-types";
 import Link from "next/link";
+import { filter } from "lodash";
 import css from "./viewerStyles.scss";
 
 const View = ({
@@ -9,6 +10,42 @@ const View = ({
   issuedStatus,
   notRevokedStatus
 }) => {
+  const errorMessages = [
+    {
+      title: "Certificate from unregistered institution",
+      message:
+        "The institution that issued this certificate is unknown. Your institution has to register with OpenCerts first.",
+      error: !issuerIdentityStatus.verified
+    },
+    {
+      title: "Certificate revoked",
+      message:
+        "This certificate has been revoked by your issuing institution. Please contact your issuing institution for more details.",
+      error: !notRevokedStatus.verified
+    },
+    {
+      title: "Certificate has been tampered with",
+      message:
+        "The contents of this certificate are inaccurate and have been tampered with.",
+      error: !hashStatus.verified
+    },
+    {
+      title: "Certificate not issued",
+      message:
+        "This certificate cannot be found. Please contact your issuing institution for help or issue the certificate before trying again.",
+      error: !issuedStatus.verified
+    }
+    // {
+    //   title: "Certificate store address is invalid",
+    //   message:
+    //     "Please check that you have a valid smart contract with us and a correct certificate store address before proceeding.",
+    //   error: !storeStatus.verified
+    // }
+  ];
+
+  const stack = filter(errorMessages, ["error", true]);
+  const error = stack.pop();
+
   const isWarning =
     hashStatus.verified && issuedStatus.verified && notRevokedStatus.verified;
   return (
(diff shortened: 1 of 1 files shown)


src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
+    }
+    // {
+    //   title: "Certificate store address is invalid",
+    //   message:
+    //     "Please check that you have a valid smart contract with us and a correct certificate store address before proceeding.",
+    //   error: !storeStatus.verified
+    // }
+  ];
+
+  const stack = filter(errorMessages, ["error", true]);
```
rjchow: can explore using js array filter

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
+  const error = stack.pop();
```
rjchow: leave some comments on how this thing works and point out that the priority order works bottom to top

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
+      error: !hashStatus.verified
+    },
+    {
+      title: "Certificate not issued",
+      message:
+        "This certificate cannot be found. Please contact your issuing institution for help or issue the certificate before trying again.",
+      error: !issuedStatus.verified
+    }
+    // {
+    //   title: "Certificate store address is invalid",
```
rjchow: 1

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
+      error: !issuerIdentityStatus.verified
+    },
+    {
+      title: "Certificate revoked",
+      message:
+        "This certificate has been revoked by your issuing institution. Please contact your issuing institution for more details.",
+      error: !notRevokedStatus.verified
+    },
+    {
+      title: "Certificate has been tampered with",
```
rjchow: 2

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
+      message:
+        "The contents of this certificate are inaccurate and have been tampered with.",
+      error: !hashStatus.verified
+    },
+    {
+      title: "Certificate not issued",
```
rjchow: 3

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
...
 }) => {
+  const errorMessages = [
+    {
+      title: "Certificate from unregistered institution",
+      message:
+        "The institution that issued this certificate is unknown. Your institution has to register with OpenCerts first.",
+      error: !issuerIdentityStatus.verified
+    },
+    {
+      title: "Certificate revoked",
```
rjchow: 4

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -9,6 +10,42 @@ const View = ({
   issuedStatus,
   notRevokedStatus
 }) => {
+  const errorMessages = [
+    {
```
rjchow: remove

src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -38,29 +75,10 @@ const View = ({
...
-        {!issuedStatus.verified ? (
-          <p className={css.messages}>The certificate records are not found</p>
-        ) : null}
-
-        {!notRevokedStatus.verified ? (
-          <p className={css.messages}>The certificate has been revoked</p>
-        ) : null}
-
-        {!issuerIdentityStatus.verified ? (
+        {error !== null ? (
```
rjchow: check against undefined instead of null

samuellye98:
https://github.com/OpenCerts/opencerts-website/pull/357

PR #361, opened 2019-07-12
Fix(Reducers): Disable setting activeTemplateTab when registering templates

samuellye98:
In reducers/certificate, remove the activeTemplateTab: 0 from CERTIFICATE_TEMPLATE_REGISTER. This is so that when we updateTemplateTabs, activeTab will not keep switching back to 0.

### DIFF
--- a/src/reducers/certificate.test.js
+++ b/src/reducers/certificate.test.js
@@ -56,8 +56,7 @@ describe("reducers", () => {
             id: "transcript",
             label: "Transcript"
           }
-        ],
-        activeTemplateTab: 0
+        ]
       };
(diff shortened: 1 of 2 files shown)


src/reducers/certificate.test.js
```diff
@@ -56,8 +56,7 @@ describe("reducers", () => {
             id: "transcript",
             label: "Transcript"
           }
-        ],
-        activeTemplateTab: 0
```
yehjxraymond: Can you in the same commit default the activetemplatetab to 0 when a new cert is loaded?

PR #362, opened 2019-07-12
Feat/dns verification

sumitnitsurat:
- DNS verification added for document store
- Added unit test for dns verification 
- Create New issue for tls library used in dnsprove


package.json
```diff
@@ -52,6 +53,7 @@
     "redux": "^4.0.0",
     "redux-devtools-extension": "^2.13.7",
     "redux-saga": "^0.16.0",
+    "tls": "0.0.1",
```
rjchow: what's this library used for?

package.json
```diff
@@ -40,6 +40,7 @@
     "next-seo": "^1.11.2",
     "node-sass": "^4.12.0",
     "number-to-words": "^1.2.4",
+    "opencerts-dnsprove": "^2.0.0",
```
yehjxraymond: Needs to be in @govtechsg org.

package.json
yehjxraymond: If it's suppose to be in dnsprove, shoudl move there instead.

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
```diff
@@ -56,7 +57,25 @@ const renderIcon = status => {
   );
 };
 
-const renderText = status => {
+export const getIdentityVerificationText = identityStatus => {
+  const verifiedText = identityStatus.reduce(
+    (prev, next) => {
+      /* eslint-disable-next-line no-param-reassign */
+      if (next.registry && !prev.isRegistry) prev.isRegistry = true;
```
Nebulis: minor comment but the condition `!prev.isRegistry` is not necessary

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
```diff
@@ -0,0 +1,30 @@
...
+    const testValue = [{ registry: "Govtech", dns: "abc.com" }];
+    expect(getIdentityVerificationText(testValue)).toEqual("Accredited by SSG");
+  });
+
+  test("when registry is unverified but dns is verified", () => {
+    const testValue = [{ registry: false, dns: "abc.com" }];
+    expect(getIdentityVerificationText(testValue)).toEqual("Issued by ABC.COM");
+  });
+
+  test("when multiple registry is verified", () => {
```
Nebulis: let's differentiate the test case in their description

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+  test("when multiple dns is verified", () => {
+    const testValue = [
+      { registry: false, dns: "abc.com" },
+      { registry: false, dns: "demo.com" }
+    ];
+    expect(getIdentityVerificationText(testValue)).toEqual(
+      "Issued by ABC.COM & DEMO.COM"
+    );
+  });
```
Nebulis: What about those use cases:
- testValue = []
- testValue = [{ registry: "demo", dns: "demo.com" }, { registry: false, dns: "abc.com" }]
- testValue = [{ registry: false, dns: "abc.com" }, { registry: "demo", dns: "demo.com" }]

Nebulis: and so ?it's still possible to call that function that way

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
Nebulis: test description must reflect what the test is ding, there are the same description twice

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
Nebulis: up to the opener to resolve discussion :)

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
sumitnitsurat: something like this "should return appropriate display text when multiple registry is verified"

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
sumitnitsurat: okay :p

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
sumitnitsurat: when there is no registry and dns saga test will fail. this component code will never execute.

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
Nebulis: yup this sounds better

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
Nebulis: What's the issue on setting isRegistry again and again ?

Anyway I actually have another suggestion to change the function:

```
if(some(identityStatus, (idontknowhowtonamethis) => idontknowhowtonamethis.isRegistry)) {
  return "Accredited by SSG";
} else {
// note filter Boolean is to remove empty values
  return `Issued by ${identityStatus.map((idontknowhowtonamethis) =>idontknowhowtonamethis.dns).filter(Boolean).dns.join(" & ")}`
}
```

the issue with the current function is that there is `no comments/explanation` and the intent using a reducer is not really clear as it's doing 2 things at the same time

rjchow (changes requested):
the verification logic looks correct but the flow is hard to follow - we'll need to refactor this in the following days.

aside from that there's also the rendering issue which i posted a screenshot of on telegram

do add in the integration test cases below as well for the various combinations

### Single Issuer
##### Valid (Green)
######Schema 1.5:
Accredited Issuer

###### Schema 2.0
Accredited Issuer
DNS Issuer

##### Invalid (means it should show red and not progress)
###### Schema 1.5
Issuer not on registry
###### Schema 2.0
No identityProof, issuer not on registry
Has identityProof, DNS records has wrong address
Has identityProof, no OpenAtts records

### Multiple Issuer
##### Valid
All accredited issuers: should show accredited cert
All DNS issuers: should show issued by DNS names
Mix of accredited issuers & DNS issuers: should show accredited cert

##### Invalid (show red and not render cert)
Mix of valid and invalid issuers
All invalid issuers

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
sumitnitsurat: We are already iterating on the same array twice in our code ..with this method in case of dns it will iterate 4 times on the same array to return us the required value..

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
sumitnitsurat: updated please check .

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
Nebulis: What's the issue with that ? performance concern to iterate over tiny array ?

readabiliy > performance unless performance is an actual issue. On top of that this function iterate only twice in that case

sumitnitsurat: ok

src/sagas/certificate.js
```diff
@@ -253,94 +247,144 @@ function isApprovedENSDomain(issuerAddress) {
...
-    const contractStoreAddresses = get(data, "issuers", []).map(issuer =>
-      getDocumentStore(issuer)
+export function* verifyCertificateDnsIssuer({ issuer }) {
+  const location = get(issuer, "identityProof.location");
+
+  if (!location) return false;
+
+  const dnsRecords = yield call(getDocumentStoreRecords, location);
+
+  trace(`DNS records: ${dnsRecords}`); // dnsRecords: [{addr: "0xabc", netId: 3}]
```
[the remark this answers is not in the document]
sumitnitsurat: fixed it.

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
sumitnitsurat: Should I resolve this ?

src/components/CertificateVerifyBlock/CertificateVerifyBlock.js
sumitnitsurat: fixed it ..please check

src/sagas/certificate.test.js
```diff
@@ -77,37 +94,42 @@ function whenThereIsOnlyOneEnsName() {
...
+function whenThereAreDnsNameAndEthereumAddress() {
+  const dnsName = "abc.com";
+  const ethereumAddress = "0x0c9d5E6C766030cc6f0f49951D275Ad0701F81E2";
   const testCert = new MakeCertUtil()
-    .addIssuer(ensNames[0])
-    .addIssuer(ensNames[1])
+    .addDataField("issuers", [
+      {
+        identityProof: { type: "DNS-TXT", location: dnsName },
+        documentSotre: ethereumAddress
```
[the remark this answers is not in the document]
sumitnitsurat: fixed it.

sumitnitsurat:
> the verification logic looks correct but the flow is hard to follow - we'll need to refactor this in the following days.
> 
> aside from that there's also the rendering issue which i posted a screenshot of on telegram
> 
> do add in the integration test cases below as well for the various combinations
> 
> ### Single Issuer
> ##### Valid (Green)
> ######Schema 1.5:
> Accredited Issuer
> 
> ###### Schema 2.0
> Accredited Issuer
> DNS Issuer
> 
> ##### Invalid (means it should show red and not progress)
> ###### Schema 1.5
> Issuer not on registry
> 
> ###### Schema 2.0
> No identityProof, issuer not on registry
> Has identityProof, DNS records has wrong address
> Has identityProof, no OpenAtts records
> 
> ### Multiple Issuer
> ##### Valid
> All accredited issuers: should show accredited cert
> All DNS issuers: should show issued by DNS names
> Mix of accredited issuers & DNS issuers: should show accredited cert
> 
> ##### Invalid (show red and not render cert)
> Mix of valid and invalid issuers
> All invalid issuers

I will create a new issue for that to refactor and add integration test for this.

rjchow:
do the tests in this merge so that we can have confidence that the verification logic works - right now we don't run any tests to check that cert verification is correct so that's very dangerous

rjchow (commented):
left two comments the rest are correct

rjchow (approved):

PR #364, opened 2019-07-15
Fix(Templates): Enable default template viewing

samuellye98:
Change CertificateViewer.js logic to render default template from legacy.opencerts.io if $template is missing.

### DIFF
--- a/src/components/CertificateViewer.js
+++ b/src/components/CertificateViewer.js
@@ -94,9 +94,9 @@ const CertificateViewer = props => {
       <DecentralisedRenderer
         certificate={document}
         source={`${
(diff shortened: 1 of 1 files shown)


src/components/CertificateViewer.js
```diff
@@ -94,9 +94,9 @@ const CertificateViewer = props => {
       <DecentralisedRenderer
         certificate={document}
         source={`${
-          typeof document.data.$template === "string"
-            ? LEGACY_OPENCERTS_RENDERER
-            : certificate.$template.url
+          typeof document.data.$template === "object"
```
rjchow: typeof null is object btw

Issue #359, opened 2019-07-09
feat(loading): Loading background while waiting for decentralised renderer to respond


yehjxraymond:
Add a background with the loading icon to show users that the template is loading. 

As we won't know when the iframe has loaded, we will just add that to the background so that when the iframe loads, it will cover the background and hide the loader.

thefiend:
Added in #369 by @samuellye98 .

PR #369, opened 2019-07-16
Feat(loading): Loading background while waiting for decentralised renderer

samuellye98:
https://github.com/OpenCerts/opencerts-website/issues/359


src/components/CertificateViewer.js
```diff
@@ -28,6 +28,15 @@ const renderVerifyBlock = props => (
   />
 );
 
+const LoadingIframe = (
```
yehjxraymond:
```suggestion
const LoadingIframe = () =>
```

src/components/CertificateViewer.js
```diff
@@ -91,14 +100,17 @@ const CertificateViewer = props => {
...
-      <DecentralisedRenderer
-        certificate={document}
-        source={`${
-          typeof document.data.$template === "object"
-            ? certificate.$template.url
-            : LEGACY_OPENCERTS_RENDERER
-        }`}
-      />
+      <div>
+        {LoadingIframe}
```
yehjxraymond:
```suggestion
        <LoadingIframe/>
```

src/components/CertificateViewer.js
Nebulis: I would suggest to use prettier :D

src/components/CertificateViewer.js
```diff
@@ -91,14 +100,17 @@ const CertificateViewer = props => {
...
-        certificate={document}
-        source={`${
-          typeof document.data.$template === "object"
-            ? certificate.$template.url
-            : LEGACY_OPENCERTS_RENDERER
-        }`}
-      />
+      <div>
+        <LoadingIframe />
+        <DecentralisedRenderer
```
Nebulis: What do you think if the renderer send a signal saying it's loaded ?

(to do in another MR)

samuellye98: like a notification?

Nebulis: yeah for instance, like the setHeight but just for the status, maybe instead of setHeight we should switch to an action dispatcher a la redux =)

Nebulis: Current code is (in document-renderer)

```
window.connection = window.Penpal.connectToChild({
      iframe,
      methods: {
        updateHeight,
        updateTemplates 
      }
    });
```

Suggestion is (sth like)

```
const actions = function(action) {
      if (action.type === 'SET_HEIGHT') {
        updateHeight(action.payload);
      } else if (action.type === 'SET_TEMPLATES') {
        updateTemplates(action.payload);
      } else {
        console.log(`I dont handle ${action.type}`)
      }
    };
    window.connection = window.Penpal.connectToChild({
      iframe,
      methods: {
        actions
      }
    });
```

Again out f scope of this MR =)

Nebulis (approved):

Issue #375, opened 2019-07-17
fix(print): printed cert is truncated on some environments


rjchow:
for printed demo cert:

Windows:
❓  Edge (could not drag demo cert)
✅  Firefox
❌  chrome

Linux
❌  Firefox
❌  Chrome

Mac OS X
❓  Firefox
❓  Chrome

Example:

[a.pdf](https://github.com/OpenCerts/opencerts-website/files/3399937/a.pdf)

samuellye98:
❓ Edge (could not drag demo cert)
✅ Firefox
❌ chrome

Linux
❌ Firefox
❌ Chrome

Mac OS X
✅ Firefox
✅ Chrome

Nebulis:
I've been able to reproduce by playing with the rendering from chrome (otherwise by default it looks fine for me)

ctrl (cmd) + shift + p => rendering => emulate css media => print

I had to zoom out (33%) a bit to get the following

![image](https://user-images.githubusercontent.com/503185/61352661-85879600-a8a0-11e9-9f76-e25231246de1.png)

PR #378, opened 2019-07-17
fix: #375 - test

### LINKED ISSUE #376: Bug(CSS): Distorted UI due to overriden CSS
The styles in privacy.scss is overriding the global styles, instead we need to localise scss styles to its respective component.

Refer to #373 for an example

Nebulis:
trying to fix #376. Some note:

The issue is probably due to the height of iframe being too short to display the whole content: 
- Using background-image has no impact on parent height
- No divs are sized resulting in no height computation


src/components/certificateViewer.scss
```diff
@@ -5,6 +5,10 @@
   #top-header-ui {
     display: none;
   }
+
+  iframe {
+    height: 99vh; /* dont use 100vh otherwise a second blank page appear when printing */
```
yehjxraymond: Why then 99vh?

Nebulis: because it works, there is probably like 1px causing the issue, dont want to investigate that much considering that I'm pretty sure it's not working with more than one file

rjchow (dismissed):

rjchow (approved):

PR #379, opened 2019-07-18
Chore(img): Reduce DropZone image file size

samuellye98:
Replace dropzone image with html text


src/components/HomePageContent/DropZoneSection.js
```diff
@@ -44,7 +44,15 @@ const DraggableDemoCertificate = () => (
       </div>
       <div className="col">
         <img
-          src="/static/images/dropzone/arrow3.png"
+          src="/static/images/dropzone/arrow.png"
+          width="100%"
+          draggable="false"
+        />
+        <h8 style={{ fontSize: 14, textAlign: "justify", color: "#ff9933" }}>
```
Nebulis: h8 dont exist in native browser tag :D

Rendering changed quite a lot
from 
image

to
image

Nebulis:
Icon size can probably be lowered more, however I dunno if it's really a concern:


This is the size of the loading site
image

src/components/HomePageContent/DropZoneSection.js
```diff
@@ -44,7 +44,22 @@ const DraggableDemoCertificate = () => (
...
       <div className="col">
         <img
-          src="/static/images/dropzone/arrow3.png"
+          src="/static/images/dropzone/arrow.png"
+          width="100%"
+          draggable="false"
+        />
+        <span
+          style={{
+            // fontSize: "15px",
```
Nebulis: to remove :p

thefiend:
Probably want to replace the "/static/images/dropzone/opencerts.png" with "/static/images/opencertslogo.svg" so that there is one less image and the svg image is cleaner.

Screenshot 2019-07-25 at 12 57 09 AM

PR #399, opened 2019-08-14
feat(certificate): Share certificate by link and QR code

thefiend:
Added sharing of certificate by link and QR code.

**Test:**
- Drag and drop demo certificate in home page

- Click on the link icon on the top right of the tools bar
- Click the "Copy" button
- Click the "Regenerate Link" button
- Copy the generated link and replace "https://opencerts.io" with "http://localhost:3000"


Nebulis:
What's the purpose of regenerate link ?

Nebulis:
I tried to generate a link and access to it but nothing happens. Here is the link
https://opencerts.io/?documentId=7de3bce4-de62-4628-914e-97d41e642582#aa57eb519fd3c63c42c2f2697e8957198b56fc945c4db18b480c07d2e6485a93

thefiend:
> I tried to generate a link and access to it but nothing happens. Here is the link
> https://opencerts.io/?documentId=7de3bce4-de62-4628-914e-97d41e642582#aa57eb519fd3c63c42c2f2697e8957198b56fc945c4db18b480c07d2e6485a93

You need to replace the "https://opencerts.io" with "localhost:3000".

thefiend:
> What's the purpose of regenerate link ?

The link expires after a stated duration, it allows the user to manually trigger to obtain a new link.

package.json
```diff
@@ -47,6 +49,7 @@
     "number-to-words": "^1.2.4",
     "penpal": "^4.0.0",
     "prop-types": "^15.6.1",
+    "qrcode.react": "^0.9.3",
```
rjchow: use the existing react-qr-svg library

rjchow:
Peek 2019-08-28 10-55

src/components/CertificateShareLink/CertificateShareLinkForm.js
```diff
@@ -0,0 +1,114 @@
...
+      copiedLink,
+      handleCopyLink,
+      handleShareLinkToggle,
+      generateShareLink
+    } = this.props;
+
+    const certificateLink =
+      shareLink && `${window.location.origin}/${shareLink.id}#${shareLink.key}`;
+    return (
+      <div className="container">
```
Nebulis: Could you clarify this story, I'm a bit confused, there is no detail on it
- how the link is store
- how the link is deleted once expired
- why is the link onl usable once => and so what's the point of having both regeneration and usable once
- what's the goal we are trying to achieve with the regenerate button => why is this button needed (usecase)
- anything that can be useful for people with no context on this story ..

rjchow:
image

you can't copy/paste from a disabled textbox in firefox

rjchow:
image

POST failure isn't handled

Nebulis:
Maybe we should directly make the link clickable so that it's directly in the clipboard

thefiend:
> image
> 
> you can't copy/paste from a disabled textbox in firefox


Which version  is this? the latest has copy link button.

src/components/CertificateShareLink/CertificateShareLinkForm.js
thefiend: - **how the link is store**
The unique link is created using the API provided by Raymond where it will upload the certificate using a POST method and return the document ID and encryption key in the response.

- **how the link is deleted once expired**
The link will be automatically deleted once it is expired based on the time-to-live (ttl) time set in the payload for the POST method.

- **why is the link onl usable once => and so what's the point of having both regeneration and usable once**
As discussed with the rest, I think it makes more sense to give like a 14 days or so ttl to the link so that user can send that link to their employer so that they can down the file from there and it will automatically be disabled after the HR personnel has finished reviewing the application which usually takes 2 weeks or so (maybe longer?) without the user having to keep track of every link.

The initial idea was to allow a number of visits to the link before it is disabled but that might be an issue also.

- **what's the goal we are trying to achieve with the regenerate button => why is this button needed (usecase)**
The regenerate button was added to allow the user to regenerate a new URL if the old one somehow doesn't work so that they do not have to start over again just to get a URL.

- **anything that can be useful for people with no context on this story ..**
The aim of this feature is to allow the certificate owner to share the unique link to their certificate to allow them to easily share their certificate without having to send their actual file over.

thefiend:
> image
> 
> POST failure isn't handled

Added a error screen to manage this.

rjchow:
- [ ] check that it works with a big file (10mb)

package.json
Nebulis: done

Nebulis:
@rjchow 

- removed regenerate link
- did some ui changes on the button above the viewer
- added feature flag:
  - Create a new bucket in S3 (opencerts.io)
  - Added a file publicly available with CORS authorization => https://s3-ap-southeast-1.amazonaws.com/opencerts.io/feature-toggle.json
  - Added a new feature toggle: `SHARE_LINK`
  - Created a component to ease toggle usage: `FeatureToggle`
  - New configuration variable created to handle feature toggles: `ENVIRONMENT`. The computation is made based on environment variables provided by [netlify](https://www.netlify.com/docs/continuous-deployment/?_ga=2.254249672.1986722564.1569467860-817711885.1562657089#build-environment-variables). Assuming that production website is setting `context` to `production`. Any other environment will default to development.

src/sagas/certificate.js
```diff
@@ -442,6 +444,65 @@ export function* sendCertificate({ payload }) {
...
+    });
+  } catch (e) {
+    yield put({
+      type: types.GENERATE_SHARE_LINK_FAILURE,
+      payload: e.message
+    });
+  }
+}
+
+export function* retrieveCertificateByLink({ payload }) {
```
rjchow: this does a bit more than retrieve, it actually retrieves and populates the store with the cert
i'm not sure might want to split it out or rename this

also i think "ByLink" is inaccurate since it's really retrieving it from a preconfigured address and appending documentId
maybe FromStore({id, decryptionKey}) is more accurate.

src/sagas/certificate.js
```diff
@@ -511,6 +572,8 @@ export function* analyticsStoreFail() {
 export default [
   takeEvery(types.UPDATE_CERTIFICATE, verifyCertificate),
   takeEvery(types.SENDING_CERTIFICATE, sendCertificate),
+  takeEvery(types.GENERATE_SHARE_LINK, generateShareLink),
+  takeEvery(types.GET_CERTIFICATE_BY_ID, retrieveCertificateByLink),
```
rjchow: 🤣

package.json
```diff
@@ -27,8 +27,10 @@
   "dependencies": {
     "@govtechsg/dnsprove": "^2.0.3",
     "@govtechsg/open-attestation": "^1.1.30",
+    "@govtechsg/opencerts-encryption": "^1.0.0",
```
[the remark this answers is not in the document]
Nebulis: locked

package.json
```diff
@@ -39,16 +41,17 @@
     "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.15",
     "moment-timezone": "^0.5.23",
-    "next": "^9.0.3",
+    "next": "^7.0.1",
```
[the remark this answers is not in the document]
Nebulis: oopsie while merge conflicts => reverted

package.json
```diff
@@ -39,16 +41,17 @@
...
     "next-seo": "^1.11.2",
     "node-sass": "^4.12.0",
     "number-to-words": "^1.2.4",
     "penpal": "^4.0.0",
     "prop-types": "^15.6.1",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
+    "query-string": "^6.8.2",
+    "react": "^16.6.3",
+    "react-dom": "^16.6.3",
```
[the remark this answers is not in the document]
Nebulis: oopsie while merge conflicts => reverted

(this one was ok because of ^)

pages/_app.js
```diff
@@ -4,8 +4,28 @@ import Router from "next/router";
...
 import initStore from "../src/store";
-import { GA_ID } from "../src/config";
+import { GA_ID, ENVIRONMENT } from "../src/config";
+import { types } from "../src/reducers/featureToggle";
+
+const FeatureFlagLoader = ({ dispatch, children }) => {
+  React.useEffect(() => {
+    const run = async () => {
+      const featureToggle = await fetch(
+        "https://s3-ap-southeast-1.amazonaws.com/opencerts.io/feature-toggle.json",
```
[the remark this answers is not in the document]
Nebulis: nah it's ok wont be used anywhere else (used once only), I'm ok leaving that here, unless you think it's an issue

pages/_app.js
```diff
@@ -23,7 +43,9 @@ class MyApp extends App {
     return (
       <Container>
         <Provider store={store}>
-          <Component {...pageProps} />
+          <FeatureFlagLoader dispatch={store.dispatch}>
```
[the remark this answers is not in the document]
Nebulis: as discussed it doesnt  :see_no_evil:

PR #403, opened 2019-08-15
fix(seo): Added sitemap

thefiend:
Added command and sitemap generating script based on pages

To view it, please go to:
[SITE URL]/static/sitemap.xml


package.json
```diff
@@ -11,6 +11,8 @@
     "build": "next build",
     "build:static": "next build && next export",
```
rjchow: using `npm run build` here didn't trigger postbuild?

rjchow:
sitemap.xml isn't at the root, its at `https://deploy-preview-403--opencerts-production.netlify.com/static/sitemap.xml` - is that ok?

package.json
thefiend: it gives an error because "next export" is running in parallel with "node postbuild" so I had to create a " postbuild:static"

alternatively, probably can put "next export" in "postbuild"

thefiend:
> sitemap.xml isn't at the root, its at `https://deploy-preview-403--opencerts-production.netlify.com/static/sitemap.xml` - is that ok?

discussed with Raymond, its fine, will just have to specify it in search console

PR #415, opened 2019-09-11
fix(dropbox): Dropbox not showing error for certificate with malformed type  

### LINKED ISSUE #414: Error in malformed certificate not throwing visually
[malformed.txt](https://github.com/OpenCerts/opencerts-website/files/3594445/malformed.txt)
The above cert does not throw an error when viewed. 

The problem was that the type definition for the value is not valid and the error was not thrown.  
```
    "id": "364999b4-486d-443e-96b8-3aae3eb16794:strig:53b75bbe",
```

yehjxraymond:
This fixes #414.

The problem was that in previous error handling method, we are sending an action with `certificate: getData(rawCertificate)`. The problem is when `getData` fail, the error handling function throws instead and the action for the error is not dispatched. 

The fix is to remove all the certificate in the error handling function since there is no use of it being sent in the action anyway. Also in the same PR, refactored the analytics handling saga and extracted common certificate methods to it's own utils.

### DIFF
--- a/src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
+++ b/src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
@@ -12,8 +12,11 @@ import {
   selectTemplateTab as selectTemplateTabAction
 } from "../../reducers/certificate";
 import { analyticsEvent } from "../Analytics";
-import { getAnalyticsStores } from "../../sagas/certificate";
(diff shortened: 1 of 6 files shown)


yehjxraymond:
[malformed.txt](https://github.com/OpenCerts/opencerts-website/files/3600045/malformed.txt)

This file should show error instead of showing the loading screen.

Nebulis (approved):

PR #428, opened 2019-10-11
feat: separate registry check from certificate checks

Nebulis:
## Goal

image

### Business rules
1. For the green/gray header
- if one of the issuer is in the registry => display the green header (in the registry)
- if all issuers are not in the registry => display the gray header (*NOT* in the registry)

2. For the Certificate checks
- if one of the issuer is in the registry display the first available issuer name
- if all issuers are not in the registry, display the first available name from dns resolution. If all dns resolution fails, displays unknown

### Changes from the image
The link in case of not in the registry must be *What does this mean ?*

## Help me
In the code, `CertificateVerifyBlock` and `DetailedCertificateVerifyBlock` checks for the validity of many status, but from my understanding we dont display the certificate in most of the cases (all ?), should I rake the opportunity to remove everything that is unecessary ?


src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
```diff
@@ -22,7 +22,7 @@ describe("Certificate verify block getIdentityVerificationText", () => {
         { registry: "demo", dns: "demo.com" }
       ];
       expect(getIdentityVerificationText(testValue)).toEqual(
-        "Accredited by SSG"
+        "Certificate issued by govtech"
```
sumitnitsurat: I think its better if we only have the capitalized name of the organization.

sumitnitsurat:
Screenshot 2019-10-14 at 10 49 21


"Ropsten" is coming with the name of the institute. Is it intentionally done?

Nebulis:
It's the name as available in registry.json. UI looks weird on your screenshot :see_no_evil:

Nebulis:
My bad it looks ok =)

src/components/CertificateVerifyBlock/CertificateVerifyBlock.test.js
Nebulis: you mean displaying `Govtech` or `GOVTECH` in that case ?

sumitnitsurat: yes. either one. In the previous implementation, we went with all capital.

Nebulis: I dunno I'm not fond of transforming data they set themself :thinking: I will uppercase to keep the behaviour like before

Nebulis: done

src/components/CertificateViewer.js
```diff
@@ -101,15 +103,44 @@ const renderHeaderBlock = props => {
   );
 };
 
-const CertificateViewer = props => {
+export const StatelessCertificateViewer = props => {
```
sumitnitsurat: I dont think naming them stateless is good idea.

Nebulis: it is and there is already a Certificate viewer component, I dont want to refactor this now :)

PR #429, opened 2019-10-23
feat: pass down raw document to decentralized renderer

Nebulis:
Provide raw document to decentralized renderer.

We need it for acra so that acra renderer can perform extra check regarding the target hash of the document

### DIFF
--- a/src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
+++ b/src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
@@ -49,7 +49,8 @@ class DecentralisedRenderer extends Component {
   async renderDocument(certificate) {
     const { childFrameConnection } = this.state;
     const child = await childFrameConnection;
-    await child.renderDocument(certificate);
+    const rawDocument = this.props.certificate;
+    await child.renderDocument(certificate, rawDocument);
   }
 
   // Do not re-render component if only activeTab changes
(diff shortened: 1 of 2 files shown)


src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
```diff
@@ -49,7 +49,7 @@ class DecentralisedRenderer extends Component {
   async renderDocument(certificate) {
     const { childFrameConnection } = this.state;
     const child = await childFrameConnection;
-    await child.renderDocument(certificate);
+    await child.renderDocument(certificate, this.props.certificate);
```
yehjxraymond: Might be unclear to new dev what's the difference. Maybe can rename this.props.certificate? 
```
const rawDocument = this.props.certificate
```

Nebulis: actually I dunno the diff between document and certificate in props ...

Nebulis: done

src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.js
```diff
@@ -80,7 +82,10 @@ class DecentralisedRenderer extends Component {
     this.setState({ childFrameConnection });
 
     childFrameConnection.then(frame =>
-      frame.renderDocument(getData(this.props.certificate))
+      frame.renderDocument(
+        getData(this.props.certificate),
```
yehjxraymond:
```suggestion
        getData(rawDocument),
```

PR #442, opened 2019-11-08
feat: print from frame if available

Nebulis:
Check #441 to see the difference in behavior with demo certificate when printing (try with background graphics).

Please also note that if the renderer can't print itself, it will default to app printing => try it out with TT demo document.

- [ ] is it possible to force background graphics ?
- [x] add watermark in demo renderer

### DIFF
--- a/src/components/CertificateViewer.js
+++ b/src/components/CertificateViewer.js
@@ -1,4 +1,5 @@
 import PropTypes from "prop-types";
+import React from "react";
 import dynamic from "next/dynamic";
 import Link from "next/link";
 import { get, some } from "lodash";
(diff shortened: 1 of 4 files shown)


src/components/CertificateViewer.js
```diff
@@ -41,7 +49,9 @@ const renderHeaderBlock = props => {
             <div
               id="btn-print"
               className={styles["print-btn"]}
-              onClick={() => window.print()}
+              onClick={() => {
```
yehjxraymond:
```suggestion
              onClick={childRef.current.print}
```

Nebulis: Doesnt work, on first render print is undefined and as a ref is a mutable value the component is not re-renderer when it's updated

src/components/CertificateViewer.js
```diff
@@ -131,7 +142,10 @@ export const CertificateViewer = props => {
       <div id={styles["top-header-ui"]}>
         <div className={styles["header-container"]}>{renderedHeaderBlock}</div>
       </div>
-      <DecentralisedRenderer rawDocument={document} />
+      <ForwardedRefDecentralisedRenderer
```
yehjxraymond: Question is why we cannot do like
```
<DecentralisedRenderer
 ref={childRef}
rawDocument={document}
```

Since `ForwardedRefDecentralisedRenderer` doesnt seemed to add additional logic?

PR #468, opened 2019-12-17
Add knowledge catalyst to collaborate

asepem7:
add knowledge catalyst to collaborate


asepem7:
Hi, I want to edit the description field on collaborate but I can't push can help?

PR #472, opened 2019-12-18
feat(certificate viewer): create floating button bar

simboonlong:
feat: add basic transitions to buttons
feat: smaller share buttons when in mobile breakpoint

### DIFF
--- a/src/components/CertificateViewer.js
+++ b/src/components/CertificateViewer.js
@@ -45,57 +45,62 @@ const renderHeaderBlock = (props, childRef) => {
       <div className="row">
         <div>{renderedVerifyBlock}</div>
         <div className={`${styles["share-buttons"]}`}>
-          <div className={`row flex-nowrap`}>
-            <div className="col-auto">
-              <div
-                id="btn-print"
-                className={styles["print-btn"]}
-                onClick={() => {
-                  childRef.current.print();
-                }}
-              >
-                <i className="fas fa-print" />
-              </div>
-            </div>
-            <FeatureFlagContainer
-              name="SHARE_LINK"
-              render={() => (
-                <div
-                  className="col-auto"
-                  onClick={() => props.handleShareLinkToggle()}
-                >
-                  <div id="btn-link" className={styles["send-btn"]}>
-                    <i className="fas fa-link" />
+          <div className={`${styles["share-buttons-wrap"]}`}>
+            <div className={`${styles["share-buttons-content"]}`}>
+              <div className={`row flex-nowrap`}>
+                <div className="col-auto">
+                  <div
+                    id="btn-print"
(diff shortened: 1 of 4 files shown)


src/components/certificateViewer.scss
```diff
@@ -143,6 +157,13 @@
     height: 100%;
     width: 100%;
   }
+
+  @include media-breakpoint-custom($max: 767px) {
```
Nebulis: why not just max-width ?

sebastianquek (commented):
I've got 3 comments regarding the sizing and alignment:
- The bar should be aligned with the content container, right now it's floating and feels out of place
pic


- The bar is very large with reference to other elements on the page and draws a lot of attention (given that it's already a stickied element, we don't need it to be too large), perhaps something smaller will be sufficient.

- On mobile it also floats a little too high above the bottom, it can be aligned with the content container too.

src/components/certificateViewer.scss
simboonlong: used @media max-width instead. leaving the custom breakpoint mixin in variables.scss, in case if next time need to handle edge cases for in between min + max scenarios.

src/components/CertificateViewer.js
```diff
@@ -44,57 +44,59 @@ const renderHeaderBlock = (props, childRef) => {
...
-                className={styles["send-btn"]}
-                title="Download"
+              <div id="btn-email" className={styles["send-btn"]}>
+                <i className="fas fa-envelope" />
+              </div>
+            </div>
+            <div className="col-auto">
+              <a
+                download={`${props.certificate.id}.opencert`}
+                target="_black"
```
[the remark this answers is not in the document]
simboonlong: aligned the bar, make the buttons smaller, corrected the typo.

src/styles/variables.scss
```diff
@@ -75,3 +112,23 @@ $max-width: 1280px;
     }
   }
 }
+
+@mixin media-breakpoint-custom ($min: 0, $max: 0) {
```
sebastianquek: I think we should remove this for now, so that the git history makes sense (i.e. we can trace when this was added and used at the same time).

src/components/certificateViewer.scss
```diff
@@ -130,19 +130,61 @@
...
 .print-btn,
 .send-btn {
   @include round-btn;
-  padding: 0.75rem;
-  margin-left: 1rem;
+  text-align: center;
+
+  width: 40px;
+  height: 40px;
+  line-height: 40px;
```
sebastianquek: Using `line-height` doesn't really center the icon perfectly 😊 , shall we use flex instead?

sebastianquek (dismissed):

simboonlong:
stale, closing

PR #480, opened 2020-01-28
feat: universal actions to load document

Nebulis:
- load certificate from url following https://github.com/Open-Attestation/adr/blob/master/universal_actions.md
- the application look for action query parameter at the root of the application
- able to load plain and encrypted document


src/components/CertificateDropZone/Views/UnverifiedView.js
```diff
@@ -48,6 +49,11 @@ const View = ({
       message:
         "Please check that you have a valid smart contract with us and a correct certificate store address before proceeding.",
       error: !storeStatus.verified
+    },
+    {
+      title: "Unable to load certificate with the provided parameters",
```
yehjxraymond: image

When an incorrect or unsupport action (ie PRESENTATION/REQUEST) is used it this message will be incorrect. 

Will be better to separate out unable to load vs unsupported actions.

src/components/tests/load-action-encrypted-certificate.spec.js
```diff
@@ -0,0 +1,167 @@
...
+  texts.reduce(
+    async (_prev, curr) => t.expect(component.textContent).contains(curr),
+    Promise.resolve()
+  );
+
+let storedId;
+let storedKey;
+
+// use a shared function to load only one document for the whole test suite
+const storeCertificate = async () => {
```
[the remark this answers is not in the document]
yehjxraymond: Feel free to use this:

uri: https://api.myjson.com/bins/1a9acm
key: 1b8c334a38f9ff96108303a4ba0cc592f1559eb24f5b48b70c9300c60a34d5e9

src/sagas/certificate.js
```diff
@@ -468,36 +469,53 @@ export function* generateShareLink() {
...
-        key: payload.encryptionKey,
-        type: "OPEN-ATTESTATION-TYPE-1"
-      })
-    );
 
-    if (!encryptedCertificate) {
-      throw new Error("Fail to retrieve certificate by id");
+    // if a key has been provided, let's assume
+    let certificate = yield window.fetch(uri).then(response => {
+      if (response.status >= 400 && response.status < 600) {
```
yehjxraymond: Will this affect redirection 3xx codes? 

IIRC there is a response.error that we can check for

src/sagas/certificate.js
```diff
@@ -468,36 +469,53 @@ export function* generateShareLink() {
...
-        iv: encryptedCertificate.document.iv,
-        key: payload.encryptionKey,
-        type: "OPEN-ATTESTATION-TYPE-1"
-      })
-    );
 
-    if (!encryptedCertificate) {
-      throw new Error("Fail to retrieve certificate by id");
+    // if a key has been provided, let's assume
+    let certificate = yield window.fetch(uri).then(response => {
```
yehjxraymond: Will it be better to abstract this as a service so we can test the two things independently?

yehjxraymond: Possibly handle the whole decryption and shit in the service too, i hate saga cause it's so freaking hard to test and the test arnt too useful.

src/components/CertificateDropZone/Views/UnverifiedView.js
Nebulis: there is already a different message for this :thinking:

src/sagas/certificate.js
Nebulis: I'm not sure what you want to test in that case
- fetch just fetch and throw in case of something wrong happens (4xx, 5xx) and I dont see the point to test just this.
- saga is not unit tested and I wont unit test it =)
- everything is covered from integration test

src/sagas/certificate.js
Nebulis: image

Nebulis: the example is also checking the status (we use isomorphic fetch)

pages/index.js
```diff
@@ -1,32 +1,33 @@
...
-      props.retrieveCertificateByLink(payload);
-    }
-  };
-
+const HomePage = props => {
+  const router = useRouter();
   useEffect(() => {
-    getCertificate();
-  }, []);
+    if (router.query.action) {
```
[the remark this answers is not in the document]
Nebulis: handled at root, with q parameter as discussed

src/components/tests/load-action-encrypted-certificate.spec.js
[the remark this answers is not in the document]
Nebulis: done

src/sagas/certificate.js
yehjxraymond: Service will handling both fetch and decryption so that the saga does have to handle the logic of when to decrypt. It can just fire off to the service the params passed to it. I can imagine when other algo are supported for decryption it might be easier to do it in a different file which we can test individually.

PR #510, opened 2020-06-15
fix: encode opencert data on download

gjj:
Issue:

- When downloading an OpenCert that contains special characters e.g. `#`, the data will be disappear. Most likely cos the browser thinks it's a fragment identifier and drops everything after.

Steps to reproduce:

1. Visit https://go.jiajian.me/mrdc-cert
1. Click on the Download button
1. Open up the downloaded OpenCert, and you should see the data being cut off (the original text contained `C#` so everything after `#` is dropped)

In this MR,

- Added `encodeURIComponent` to the download link
  - To test, visit [this link](https://dev.opencerts.io/?q={%22type%22:%22DOCUMENT%22,%22payload%22:{%22uri%22:%22https://raw.githubusercontent.com/OpenCerts/opencerts-website/1615ad54f6135a4ba3b92316ad6e8075042848fa/src/components/tests/fixture/sample-dns-verified-special-characters.json%22,%22permittedActions%22:[%22STORE%22],%22redirect%22:%22https://opencerts.io%22}}) and you should be able to download it. Some special characters are not visible on the renderer as it's embedded in the `data.consignment.someRandomTestData` key.
- Added e2e tests to test for document downloads with and without special characters (i.e. emojis, special characters, Chinese)

### DIFF
--- a/src/components/tests/download-certificate.spec.js
+++ b/src/components/tests/download-certificate.spec.js
@@ -1,3 +1,5 @@
+import fs from "fs";
+import downloadsFolder from "downloads-folder";
 import { Selector } from "testcafe";
 
 fixture("Download Certificate").page`http://localhost:3000`;
@@ -7,7 +9,8 @@ const IframeBlock = Selector("#iframe");
 const SampleTemplate = Selector("#root");
 const StatusButton = Selector("#certificate-status");
 const CertificateStatusBanner = Selector("#status-banner-container");
-const DownloadLink = Selector("#download-link");
+const DownloadLink = Selector("a").withAttribute("download");
+const DownloadButton = Selector("#btn-download");
 
 const validateTextContent = async (t, component, texts) =>
   texts.reduce(async (_prev, curr) => t.expect(component.textContent).contains(curr), Promise.resolve());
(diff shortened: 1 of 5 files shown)


src/components/tests/download-certificate.spec.js
```diff
@@ -0,0 +1,54 @@
...
+test("Sample document is downloaded correctly", async (t) => {
+  await t.setFilesToUpload("input[type=file]", [Document1]);
+
+  await validateTextContent(t, StatusButton, ["Certificate issued by EXAMPLE.OPENATTESTATION.COM"]);
+
+  // Simulate an OpenCert file download
+  const fileName = await DownloadLink.getAttribute("download");
+  const filePath = `${downloadsFolder()}\\${fileName}`;
+  await t.click(DownloadButton);
+  await t.wait(5000);
```
Nebulis, line 27: I would suggest to create a small function that check every 100ms (for instance) and throw an error after 10s (for instance)

gjj: Done!
- Added `waitForFileDownload` that checks `fs`'s `existsSync` to check every 100ms up to 10s
- Added the missing package.json that I accidentally left out yesterday

gjj:
Merging my fork into a branch on this repo. Will create a new merge request to merge into master.

PR #536, opened 2020-07-28
feat: new demo cert

gjj:
In this MR, we will update all of our existing demo OpenCerts to OpenAttestation v2.

Progress list:

- [ ] Mainnet
- [x] Ropsten
- [x] Rinkeby

Changes:

- Replaced `certificateStore` with `documentStore` and `identityProof`
- `"schema": "opencerts/v2.0"` will now be `"version": "https://schema.openattestation.com/2.0/schema.json"`
- Added `0x8Fc57204c35fb9317D91285eF52D6b892EC08cD3` to registry.json

See:

- https://docs.opencerts.io/docs/verifier
- https://openattestation.com/docs/advanced/identity-proofs


public/static/registry.json
```diff
@@ -250,6 +250,10 @@
       "name": "ROPSTEN: Government Technology Agency of Singapore (GovTech)",
```
Nebulis, line 250: remove the other unused one, make sure the cert in the gallery works

gjj: https://github.com/Open-Attestation/gallery/pull/37/files

I created an MR here. Anyway was the Gallery's OpenCerts file supposed to demo an encrypted version of the demo?

gjj: ok after doing a [github global search](https://github.com/search?q=0x8Fc57204c35fb9317D91285eF52D6b892EC08cD3&type=Code), seems like the doc store is being used everywhere. so as discussed... i think we keep it for now, unless we can change all of the testcases fixtures to use OA v2?

also, I noticed even on deploy preview (https://deploy-preview-536--opencerts-development.netlify.app/)...

`Fetch finished loading: GET "https://opencerts.io/static/registry.json".`

registry.json file seems to be called via absolute reference, not sure if it's intended? (in other words, my addition of a new ropsten addr will not be used unless pushed to master)

Nebulis: yeah can fix that later

PR #540, opened 2020-08-06
feat: error reporting in Google Analytics

gjj:
In this MR, we draw upon the following ideas to use Google Analytics as an error tracking tool:

- https://www.smashingmagazine.com/2011/10/improve-the-user-experience-by-tracking-errors/
- https://www.analyticsmania.com/post/tracking-errors-with-google-tag-manager/ (can explore in the future)

The idea is to store a stringified list of errors in CSV format, so that it can be processed on our Google Data Studio dashboards using calculated fields.

The following errors are sent:

- `CERTIFICATE_HASH` means the verifier could not match the calculated hash with the `targetHash`, which means the document has been tampered
- `UNISSUED_CERTIFICATE` means the store does not have a record of the `merkleRoot` i.e. the document has not been issued 
- `CERTIFICATE_STORE_NOT_FOUND` means the verifier could not find the store
- `REVOKED_CERTIFICATE` means the `targetHash` or the `merkleRoot` has been revoked by the store
- `ISSUER_IDENTITY` means that the document store address was not found in the DNS TXT records

To test:

- https://deploy-preview-540--opencerts-development.netlify.app/

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,6 +219,7 @@
     ]
   },
   "prettier": {
-    "printWidth": 120
+    "printWidth": 120,
+    "endOfLine": "auto"
   }
 }
--- a/package.json
+++ b/package.json
@@ -219,7 +219,6 @@
     ]
   },
   "prettier": {
-    "printWidth": 120,
-    "endOfLine": "auto"
+    "printWidth": 120
   }
 }
--- a/src/components/Analytics/analytics.test.ts
+++ b/src/components/Analytics/analytics.test.ts
@@ -26,8 +26,7 @@ describe("stringifyEvent", () => {
 describe("validateEvent", () => {
   it("throws if category is missing", () => {
     expect(() =>
-      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-      // @ts-ignore we expect this error to be thrown
+      // @ts-expect-error we expect this error to be thrown
       validateEvent({
         label: "LABEL",
       })
@@ -36,8 +35,7 @@ describe("validateEvent", () => {
 
   it("throws if action is missing", () => {
     expect(() =>
-      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-      // @ts-ignore we expect this error to be thrown
+      // @ts-expect-error we expect this error to be thrown
       validateEvent({
         category: "CATEGORY",
       })
@@ -78,8 +76,6 @@ describe("event", () => {
 
   it("sends and log ga event if window.ga is present", () => {
     const win = { ga: jest.fn() };
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
     analyticsEvent(win, evt);
     expect(win.ga).toHaveBeenCalledWith("send", "event", "TEST_CATEGORY", "TEST_ACTION", "TEST_LABEL", 2, undefined);
   });
@@ -87,8 +83,6 @@ describe("event", () => {
   it("throws if there is a validation error", () => {
     const win = { ga: jest.fn() };
     const errEvt = { ...evt, value: "STRING" };
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
     expect(() => analyticsEvent(win, errEvt)).toThrow("Value must be a number");
   });
 });
@@ -96,15 +90,13 @@ describe("event", () => {
 describe("analytics*", () => {
   // eslint-disable-next-line jest/no-hooks
   beforeEach(() => {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
+    // @ts-expect-error
     // eslint-disable-next-line jest/prefer-spy-on
     window.ga = jest.fn();
   });
   // eslint-disable-next-line jest/no-hooks
   afterEach(() => {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore
+    // @ts-expect-error
     // eslint-disable-next-line jest/prefer-spy-on
     window.ga = undefined; // This vs. delete window.ga, mockGA.mockReset()?
   });
@@ -113,8 +105,7 @@ describe("analytics*", () => {
     const mockGA = jest.fn();
     // eslint-disable-next-line jest/no-hooks
     beforeEach(() => {
-      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-      // @ts-ignore dont care about ts here
+      // @ts-expect-error dont care about ts here
       window.ga = mockGA;
     });
 
--- a/src/components/Analytics/analytics.test.ts
+++ b/src/components/Analytics/analytics.test.ts
@@ -92,107 +92,125 @@ describe("event", () => {
     expect(() => analyticsEvent(win, errEvt)).toThrow("Value must be a number");
   });
 });
-describe("sendEventCertificateViewedDetailed", () => {
-  const mockGA = jest.fn();
+
+describe("analytics*", () => {
   // eslint-disable-next-line jest/no-hooks
   beforeEach(() => {
     // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore dont care about ts here
-    window.ga = mockGA;
+    // @ts-ignore
+    // eslint-disable-next-line jest/prefer-spy-on
+    window.ga = jest.fn();
   });
-
   // eslint-disable-next-line jest/no-hooks
   afterEach(() => {
-    delete window.ga;
-    mockGA.mockReset();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
+    // eslint-disable-next-line jest/prefer-spy-on
+    window.ga = undefined; // This vs. delete window.ga, mockGA.mockReset()?
   });
 
-  describe("when is in the registry", () => {
-    it("should use document store to retrieve registry information", () => {
-      const issuer: v2.Issuer = {
-        documentStore: "0x007d40224f6562461633ccfbaffd359ebb2fc9ba",
-        name: "aaa",
-        identityProof: {
-          location: "aa.com",
-          type: v2.IdentityProofType.DNSTxt,
-        },
-      };
-      const certificateData = { id: "id1", name: "cert name", issuedOn: "a date" };
-      sendEventCertificateViewedDetailed({ issuer, certificateData });
-      expect(mockGA.mock.calls[0][0]).toStrictEqual("send");
-      expect(mockGA.mock.calls[0][1]).toStrictEqual("event");
-      expect(mockGA.mock.calls[0][2]).toStrictEqual("CERTIFICATE_DETAILS");
-      expect(mockGA.mock.calls[0][3]).toStrictEqual("VIEWED - Government Technology Agency of Singapore (GovTech)");
-      expect(mockGA.mock.calls[0][4]).toStrictEqual(
-        '"store":"0x007d40224f6562461633ccfbaffd359ebb2fc9ba";"document_id":"id1";"name":"cert name";"issued_on":"a date";"issuer_name":"Government Technology Agency of Singapore (GovTech)";"issuer_id":"govtech-registry"'
-      );
-      expect(mockGA.mock.calls[0][5]).toBeUndefined();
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -219,6 +219,7 @@
     ]
   },
   "prettier": {
-    "printWidth": 120
+    "printWidth": 120,
+    "endOfLine": "auto"
```
gjj, line 223: temporary fix for windows (it's really annoying), will remove at the end

src/components/Analytics/index.ts
```diff
@@ -47,9 +48,7 @@ export const sendEventCertificateViewedDetailed = ({
   const id = certificateData?.id ?? "";
   const name = certificateData?.name ?? "";
   const issuedOn = certificateData?.issuedOn ?? "";
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-  // @ts-ignore ignoring because typescript complain the key cant be undefined and there is no match between string type and keyof typeof registry.issuers
-  const registryIssuer = registry.issuers[store];
+  const registryIssuer = get(registry.issuers, store);
```
gjj, line 51: I think it might be better to use lodash's `get` here instead of using `@ts-ignore`

src/sagas/certificate.ts
```diff
@@ -71,6 +73,46 @@ export function* triggerAnalytics(errorCode: number) {
...
+  const id = get(certificate, "id") ?? ""; // Use get or certificate?.id?
+  const name = get(certificate, "name") ?? "";
+  const issuedOn = get(certificate, "issuedOn") ?? "";
+  const errors = value ?? "";
+
+  // If there are multiple issuers in a certificate, we send multiple events!
+  certificate.issuers.forEach((issuer: v2.Issuer) => {
+    const store = issuer.certificateStore ?? issuer.documentStore ?? issuer.tokenRegistry ?? "";
+    let issuerName = issuer?.name;
+    const registryIssuer = get(registry.issuers, store); // Instead of (a) registry.issuers[store] which causes a type error and (b) using ts-ignore
```
gjj, line 90: Using `@ts-ignore` here actually hid another error that I spent some time figuring out that I didn't import `registry.json`.......................

So I think it might be better to use lodash's `get` here instead of using `@ts-ignore`, and that's why I changed the same line in `index.ts` too.

src/sagas/certificate.ts
```diff
@@ -71,6 +73,46 @@ export function* triggerAnalytics(errorCode: number) {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function* triggerAnalyticsErrorV2(value: string) {
```
gjj, line 77: For now, we collect the same info twice.

- `triggerAnalyticsError` is the old GA error reporting code
- `triggerAnalyticsErrorV2` is the new one

Key difference is that V2 uses custom dimensions, where we store a CSV list of errors

src/sagas/certificate.ts
```diff
@@ -71,6 +73,46 @@ export function* triggerAnalytics(errorCode: number) {
...
+  const rawCertificate: WrappedDocument = yield select(getCertificate);
+  const certificate = getData(rawCertificate);
+
+  const id = get(certificate, "id") ?? ""; // Use get or certificate?.id?
```
gjj, line 81: does it matter if we use `get` vs accessing it directly?

package.json
Nebulis: What's the problem?

src/sagas/certificate.ts
Nebulis: To remove v1 and for me to check if not used in data studio

package.json
gjj: image

src/sagas/certificate.ts
Nebulis: Depends. In some case typing is still working in some cases it's inferred to any.

In that case why do we need get? With the correct type there shouldn't be any problem

package.json
Nebulis: as discussed, to remove ?

src/components/Analytics/index.ts
Nebulis: this does not solve anything as get will resolve registryIssuer as any. So at best you remove the ts-ignore.

Now that you raised it let's fix it correctly, I let you figure out. [Click this link to get some hint](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types)

src/sagas/certificate.ts
Nebulis: can remove v1

src/sagas/certificate.ts
Nebulis: so certificate is resolved as ny so both . notation and get from lodash will resolve to any too.

Let's go with dot notation as we have improved typings in a new version of oa

src/sagas/certificate.ts
Nebulis: it's not better, both wont raise error, let's fix that correctly as suggested above

src/sagas/certificate.ts
```diff
@@ -116,21 +158,30 @@ export function* verifyCertificate({ payload: certificate }: { payload: WrappedD
...
         yield call(analyticsHashFail);
+        errors.push("CERTIFICATE_HASH");
       }
       if (!isValid(fragmentsWithoutRevoke, ["DOCUMENT_STATUS"]) && certificateNotIssued(fragments)) {
         yield call(analyticsIssuedFail);
+        errors.push("UNISSUED_CERTIFICATE");
       }
       if (!isValid(fragments, ["DOCUMENT_STATUS"]) && !certificateNotIssued(fragments)) {
         yield call(analyticsStoreFail);
+        errors.push("CERTIFICATE_STORE"); // should we have a better name..? doesn't say much. maybe CERTIFICATE_STORE_NOT_FOUND?
```
Nebulis, line 172: yeah CERTIFICATE_STORE_NOT_FOUND is better

src/sagas/certificate.ts
```diff
@@ -116,21 +158,30 @@ export function* verifyCertificate({ payload: certificate }: { payload: WrappedD
...
         yield call(analyticsRevocationFail);
+        errors.push("REVOKED_CERTIFICATE");
       }
       if (!isValid(fragments, ["ISSUER_IDENTITY"])) {
         yield call(analyticsIssuedFail);
+        errors.push("ISSUER_IDENTITY");
+      }
+
+      if (errors) {
+        yield call(triggerAnalyticsErrorV2, errors.join(",")); // CSV list of all the certificate errors
```
Nebulis, line 184: just pass in the errors as an array here and let the function triggerAnalyticsErrorV2 do the transformation to csv if needed

package.json
gjj: removed! for future reference, https://stackoverflow.com/a/29888735/950462 this worked for me

src/sagas/certificate.test.ts
```diff
@@ -125,55 +118,93 @@ describe("sagas/certificate", () => {
...
+              identityProof: {
+                type: v2.IdentityProofType.DNSTxt,
+                location: "example.openattestation.com",
+              },
+            },
+          ],
+        };
+        analyticsGenerator.next(certificate);
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        // @ts-ignore
```
Nebulis, line 188: let's use ts-expect-error here

src/sagas/certificate.ts
```diff
@@ -33,72 +36,56 @@ import { getLogger } from "../utils/logger";
...
     error(e.message);
     return {};
   }
 }
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export function* triggerAnalytics(errorCode: number) {
-  const { storeAddresses, id } = yield call(getAnalyticsDetails);
-  if (storeAddresses && id) {
+export function* triggerErrorLogging(errors: string[]) {
```
Nebulis, line 53: this doesnt need to be a generator, can be a normal function

src/sagas/certificate.ts
```diff
@@ -116,21 +103,25 @@ export function* verifyCertificate({ payload: certificate }: { payload: WrappedD
...
       if (!isValid(revokeFragment, ["DOCUMENT_STATUS"])) {
-        yield call(analyticsRevocationFail);
+        errors.push("REVOKED_CERTIFICATE");
       }
       if (!isValid(fragments, ["ISSUER_IDENTITY"])) {
-        yield call(analyticsIssuedFail);
+        errors.push("ISSUER_IDENTITY");
+      }
+
+      if (errors) {
```
Nebulis, line 123: this is always true, must do, errors.length > 0

gjj: good catch! i'll use `if (errors.length)`

Nebulis: I will suggest to stick with `errors.length > 0` main reason is that
- using errors.length only is less clear on the intent vs using the comparison operator
- it can be tricky with some libs, for instance react => https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/sagas/certificate.ts
```diff
@@ -116,21 +48,25 @@ export function* verifyCertificate({ payload: certificate }: { payload: WrappedD
...
       }
       if (!isValid(revokeFragment, ["DOCUMENT_STATUS"])) {
-        yield call(analyticsRevocationFail);
+        errors.push("REVOKED_CERTIFICATE");
       }
       if (!isValid(fragments, ["ISSUER_IDENTITY"])) {
-        yield call(analyticsIssuedFail);
+        errors.push("ISSUER_IDENTITY");
+      }
+      console.log(certificate);
```
Nebulis, line 67: :eyes:

gjj: removed!

src/components/Analytics/index.ts
```diff
@@ -79,3 +77,60 @@ export const sendEventCertificateViewedDetailed = ({
...
+  const name = certificate?.name;
+  const issuedOn = certificate?.issuedOn;
+  const errorsList = errors.join(",");
+
+  // If there are multiple issuers in a certificate, we send multiple events!
+  certificate.issuers.forEach((issuer: v2.Issuer) => {
+    const store = issuer.certificateStore ?? issuer.documentStore ?? issuer.tokenRegistry ?? "";
+    let issuerName = issuer.name;
+    let registryId = null;
+    let registryIssuer: RegistryIssuer | null = null;
```
gjj, line 110: @Nebulis ok so I had to create a `RegistryIssuer` interface if not `registryIssuer.id` would not be accessible.

As for the `| null`, should I do this instead so that we don't have to use a nullable type?

```
if (isInRegistry(store)) {
  const registryIssuer: RegistryIssuer = registry.issuers[store];
  issuerName = registryIssuer.name;
  registryId = registryIssuer.id;
}
```

src/components/Analytics/analytics.test.ts
```diff
@@ -72,121 +76,119 @@ describe("event", () => {
...
+          documentStore: "0x007d40224f6562461633ccfbaffd359ebb2fc9ba",
+          name: "aaa",
+          identityProof: {
+            location: "aa.com",
+            type: v2.IdentityProofType.DNSTxt,
+          },
+        };
+        const certificateData = { id: "id1", name: "cert name", issuedOn: "a date" };
+        sendEventCertificateViewedDetailed({ issuer, certificateData });
+        expect(window.ga).toHaveBeenCalledWith(
```
gjj, line 117: Instead of using `expect(mockGA.mock.calls[0][0]).toStrictEqual("send");`, I changed it to follow `triggerErrorLogging`'s style, which is to use `expect(window.ga).toHaveBeenCalledWith(...)` so that it's consistent

src/components/Analytics/index.ts
Nebulis, line 110: this is useless here ? it's only use in the if block

src/components/Analytics/index.ts
Nebulis: so if (isInRegistry(store)) { ?

src/components/Analytics/index.ts
gjj: yep, removed since it was unnecessary. moved it to the `if` block instead

src/components/Analytics/index.ts
gjj: missed that out ps! i've since removed lodash's `get` entirely from analytics

PR #541, opened 2020-08-07
fix: logo size

gjj:
https://www.pivotaltracker.com/n/projects/2397228/stories/174207008

### DIFF
--- a/src/components/HomePageContent/partnersPage.module.scss
+++ b/src/components/HomePageContent/partnersPage.module.scss
@@ -34,9 +34,8 @@
     padding: 0 1.7rem;
   }
 
-  #SIM,
-  #LASALLE {
-    width: 50%;
+  #NYP {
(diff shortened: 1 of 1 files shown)


src/components/HomePageContent/partnersPage.module.scss
```diff
@@ -34,9 +34,8 @@
     padding: 0 1.7rem;
   }
 
-  #SIM,
-  #LASALLE {
-    width: 50%;
```
gjj, line 39: I moved them below, descending order. 😄

src/components/HomePageContent/partnersPage.module.scss
```diff
@@ -34,9 +34,8 @@
...
+  #NYP {
+    width: 95%
```
gjj, line 38: changed as per request

PR #544, opened 2020-08-17
chore: upgrade oa and oc-verify

Nebulis:
needs https://github.com/Open-Attestation/decentralized-renderer-react-components/pull/28


gjj:
https://github.com/Open-Attestation/oa-verify/pull/110

gjj:
OK so I'm done w handling a `bad response` error returned by Ethers.js (over at https://github.com/Open-Attestation/oa-verify/pull/110)

### To test locally

Here's how it looks like. If you want to test,

1. do so locally with `ETHEREUM_PROVIDER=cloudflare NET=mainnet npm run dev`
2. append `/?q={"type":"DOCUMENT","payload":{"uri":"https://gist.githubusercontent.com/gjj/1759e1627c562afbdc7a684df23c25de/raw/bbdccc871bcaae8249ad0331942e0b373445e94d/SOR_ALL_2012GCEO2376_00922019-09-02-redacted.opencert","permittedActions":["STORE"]}}`
3. spam browser's back button until you get the 502 bad gateway error

### Here's how it'd look like

Untitled

In the case of a 429 rate limit (caused by user who flood the verifier), or a 502 bad gateway (caused by Cloudflare Ethereum Gateway), we don't know for certain that "This certificate is not valid".

### My question

I'm not sure if it's conflicting to say both "certificate is not valid" and "unable to connect to the Ethereum network". People might get confused, what do yall think?

Nebulis:
@gjj yeah the label certificate is not valid is problematic here let's change that too

src/components/CertificateDropZone/Views/UnverifiedView.tsx
```diff
@@ -2,32 +2,51 @@ import { VerificationFragment } from "@govtechsg/oa-verify";
...
+      errors.splice(0, errors.length);
+      errors.push(TYPES.MISSING_RESPONSE);
+    } else if (badResponse(verificationStatus)) {
+      errors.splice(0, errors.length);
+      errors.push(TYPES.BAD_RESPONSE);
+    } else {
+      // TODO :)
+    }
+    // Somehow there are other errors that got caught by this... integration-merkle.spec.js's "odd-length" error is considered unhandled
+    // if (unhandledError(verificationStatus)) {
```
gjj, line 46: @Nebulis not sure if we should handle `unhandledError`s because I just found out another funky error classified as "unhandled".

A wrong merkle root (or rather, an odd-length merkle root) would return "Error with smart contract 0xE5c693015460868F817799CAFb363EEDb4F9E446: hex data is odd-length" and `ETHERS_UNHANDLED_ERROR`.

Nebulis: Are you sure it's related to merge root? Sounds related to contact address instead

Should be handled by verify before reaching unhandled

gjj:
For future reference, HTTP-related errors from Ethers: https://github.com/ethers-io/ethers.js/blob/9640e864a68b4a9e84e820f0ceaf1eb56c66715f/packages/web/src.ts/index.ts

package.json
```diff
@@ -22,7 +22,7 @@
     "analyze": "BUNDLE_ANALYZE=browser npm run start",
     "dev-cli": "node -r esm ./scripts/cli.js",
     "serve-static": "http-server out -s -p 3000",
-    "integration": "testcafe chrome src/**/*.spec.js",
+    "integration": "testcafe chrome src/**/integration-bad-response.spec.js",
```
Nebulis, line 25: 👀

gjj: thanks for catching that!!

src/components/tests/download-certificate.spec.js
```diff
@@ -37,7 +38,7 @@ test("Sample document is downloaded correctly", async (t) => {
   // Simulate an OpenCert file download
   const fileName = await DownloadLink.getAttribute("download");
   await t.click(DownloadButton);
-  const filePath = `${downloadsFolder()}/${fileName}`;
+  const filePath = normalize(`${downloadsFolder()}/${fileName}`); // Caters to both Windows and Linux devs
```
Nebulis, line 41: path.join instead ?https://nodejs.org/api/path.html#path_path_join_paths

src/components/Analytics/analytics.test.ts
```diff
@@ -350,5 +350,260 @@ describe("analytics*", () => {
...
+            "516794e14a69074a1437ddb972a063d04b4733ad6076fdd335f2d5bdab52ec05",
+            "74f086ebff10a10158233dbbdc207d4514f7bd8df9cbcadf4a384c1cf0de96f7",
+            "eddfd98891328db274321edc5985d839772e3bc1e481d8f7d6fa972cad7ed68f",
+          ],
+          merkleRoot: "6b5bb39220415b92c321cb981fccf6e8cdc16287d8eb2e56a2ad33f757ccfae2",
+        },
+      };
+
+      triggerErrorLogging(certificate, [
+        "SERVER_ERROR", // merkleRoot is odd-length
```
Nebulis, line 502: wrong comment

src/components/CertificateDropZone/Views/UnverifiedView.tsx
```diff
@@ -2,32 +2,51 @@ import { VerificationFragment } from "@govtechsg/oa-verify";
...
+      errors.push(TYPES.ADDRESS_INVALID);
+    } else if (contractNotFound(verificationStatus)) {
+      // if the error is because the contract cannot be found, then get rid of all errors and only keep this one
+      errors.splice(0, errors.length);
+      errors.push(TYPES.CONTRACT_NOT_FOUND);
+    } else if (serverError(verificationStatus)) {
+      // if the error is because cannot connect to Ethereum, then get rid of all errors and only keep this one
+      errors.splice(0, errors.length);
+      errors.push(TYPES.SERVER_ERROR);
+    } else if (unhandledError(verificationStatus)) {
```
Nebulis, line 43: this is a bit weird, `unhandledError` supposed to be the last error right ? so why is there a else block ? (and also we ignore this else block)

package.json
```diff
@@ -27,10 +27,10 @@
     "integration:watch": "nodemon --exec \"npm run integration\" --watch src"
   },
   "dependencies": {
-    "@govtechsg/decentralized-renderer-react-components": "^3.0.0",
+    "@govtechsg/decentralized-renderer-react-components": "^3.0.4",
     "@govtechsg/oa-encryption": "^1.3.1",
-    "@govtechsg/open-attestation": "^3.7.0",
-    "@govtechsg/opencerts-verify": "^1.2.0",
+    "@govtechsg/open-attestation": "^3.11.1",
+    "@govtechsg/opencerts-verify": "github:OpenCerts/verify#feat/new-errors",
```
Nebulis, line 33: to change before merging

src/components/CertificateDropZone/Views/UnverifiedView.tsx
```diff
@@ -2,32 +2,53 @@ import { VerificationFragment } from "@govtechsg/oa-verify";
...
+      // if the error is because the contract cannot be found, then get rid of all errors and only keep this one
+      errors.splice(0, errors.length);
+      errors.push(TYPES.CONTRACT_NOT_FOUND);
+    } else if (serverError(verificationStatus)) {
+      // if the error is because cannot connect to Ethereum, then get rid of all errors and only keep this one
+      errors.splice(0, errors.length);
+      errors.push(TYPES.SERVER_ERROR);
+    } else if (invalidArgument(verificationStatus)) {
+      // if the error is because of the merkleRoot, it should already be handled by the DOCUMENT_INTEGRITY fragment
+      // empty here so that it's not caught by ETHERS_UNHANDLED_ERROR
```
gjj, line 46: @Nebulis ok so previously I discovered that `merkleRoot`s that were invalid (odd length, incorrect length/not 64 chars, contains non-hexcadecimal values) were not handled.

Now that it's handled, we expect it to show "Certificate has been tampered with" instead, and that is handled by `DOCUMENT_INTEGRITY` instead of `DOCUMENT_STATUS`.

Not sure what you think about having this check, but nothing is in here? (It should already be handled by this line https://github.com/OpenCerts/opencerts-website/pull/544/files#diff-ad34607bf39b4b3c4414917cf9b807f8R22)

Nebulis: this is for ga right ? at least this could give us mor einformation about what really happened

src/components/CertificateDropZone/Views/UnverifiedView.tsx
```diff
@@ -46,41 +67,50 @@ interface UnverifiedViewProps {
...
-            }}
-            className={css["text-link"]}
-          >
-            Try another
-          </a>
-        </Link>
-      </span>
+      <div className={css["secondary-links"]}>
+        <span>
+          <Link href=" ">
```
Nebulis, line 101: this will reload the page isnt it ?

src/components/CertificateDropZone/Views/UnverifiedView.tsx
gjj: (explained)

src/components/CertificateDropZone/Views/UnverifiedView.tsx
gjj: changed to not use `Link`. instead, it's now a `span` element. `text-link` css class now includes `cursor: pointer`

gjj (approved):

PR #553, opened 2020-09-24
chore: container alignment

simboonlong:
### pr updates
- mostly `.container` alignment fixes (didn't like how `.main` was repeated used as "container" in many scoped styles)
- remove custom css styles if possible, use helper/bootstrap classes instead
- not gonna touch viewer page related styles yet (there's some existing weird styles, `@print` styles need to be careful also, some other day then i see)


simboonlong:
closing, will work on this instead https://github.com/OpenCerts/opencerts-website/pull/569

Issue #551, opened 2020-09-02
File picker shows up when clicking "What should I do?"


gjj:
![gif](https://i.imgur.com/6naOVkJ.gif)

a video paints a thousand pictures 😆 

unintended behaviour, will be fixed soon

PR #555, opened 2020-09-28
fix: stop file picker when clicking "What should I do?"

gjj:
This MR attempts to fix https://github.com/OpenCerts/opencerts-website/issues/551.

### DIFF
--- a/src/components/CertificateDropZone/Views/UnverifiedView.tsx
+++ b/src/components/CertificateDropZone/Views/UnverifiedView.tsx
@@ -110,12 +110,13 @@ export const UnverifiedView: React.FunctionComponent<UnverifiedViewProps> = ({ r
   );
 };
 const WhatShouldIDo: React.FunctionComponent = () => {
+  // see https://stackoverflow.com/questions/64091060/react-dropzone-prevent-inner-element-from-showing-file-picker
   const { getRootProps } = useDropzone({ noClick: true }); // doesn't work
   return (
     <Link href="/faq">
       <div
         {...getRootProps({
-          onClick: (event) => event.stopPropagation(), // this is bad for many reasons
+          onClick: (event) => event.stopPropagation(), // this is bad, but we'll use it for now until there's a fix
         })}
         className={css["unverified-btn"]}
       >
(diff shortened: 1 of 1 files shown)


src/components/CertificateDropZone/Views/UnverifiedView.tsx
```diff
@@ -110,3 +109,18 @@ export const UnverifiedView: React.FunctionComponent<UnverifiedViewProps> = ({ r
     </div>
   );
 };
+const WhatShouldIDo: React.FunctionComponent = () => {
+  const { getRootProps } = useDropzone({ noClick: true }); // doesn't work
+  return (
+    <Link href="/faq">
+      <div
+        {...getRootProps({
+          onClick: (event) => event.stopPropagation(), // this is bad for many reasons
```
gjj, line 118: I found that `stopPropagation` should be avoided (see [my question on SO](https://stackoverflow.com/questions/64091060/react-dropzone-prevent-inner-element-from-showing-file-picker)) as it may unintentionally prevent other events from happening (e.g. analytics).

Tbh I can't find out what else should I do. I tried `preventDefault` but I realise it prevents the `Link href` click instead.

gjj: @Nebulis it's not very important but if you have time, could you help to take a look? 😅 

https://stackoverflow.com/questions/64091060/react-dropzone-prevent-inner-element-from-showing-file-picker

Nebulis: I think it's fine here, events bubble from child to parent and there is nothing here. You can add the link of your SO questions in the comment :)

Nebulis (approved):

src/components/CertificateDropZone/Views/UnverifiedView.tsx
gjj: ok I added the link, will revisit once I get a reply from react-dropzone. think you have to re-approve this MR again!

Nebulis (approved):

PR #566, opened 2020-10-13
Kc collaborate

KCTechlead:
Update KC Info on collaborate


Nebulis:
this has already been merged => https://github.com/OpenCerts/opencerts-website/pull/563. On top of that there is a conflict with that change

PR #584, opened 2020-12-09
fix: ie and edge drag

simboonlong:
- to degrade gracefully for both ie and edge
- will be showing button to click on directly for sample, instead of drag drop


Nebulis:
ie or edge ?

simboonlong:
oh was trying out, didn't work

src/components/HomePageContent/DropZoneSection.tsx
```diff
@@ -51,16 +50,36 @@ const MobileDemoCertificate: React.FunctionComponent = () => (
...
+    this.state = {
+      isIeOrEdge: false,
+    };
   }
   componentDidMount(): void {
+    const browser = Bowser.getParser(window.navigator.userAgent);
+    // https://github.com/lancedikson/bowser/blob/master/src/constants.js
+    const isValidBrowser = browser.satisfies({
+      windows: {
+        chrome: "~87",
```
Nebulis, line 74: does that means that people using google chrome on windows will see the button ?

simboonlong: sadly yeah, not too good a solution eh?

Nebulis: The lib doesn't support it? 

https://stackoverflow.com/questions/56360225/how-to-detect-microsoft-chromium-edge-chredge-edgium-in-javascript

simboonlong: i thought there was an existing issue with the lib -> bowser/issues/387, as loading the demo on SOE itself was reflecting `Chrome` for both ie and edge browsers

decided to change the lib, both ie and edge browsers are still reflecting as `Chrome` when implemented. But if say i manually pasted the library into browser console and print out browser name it reflects as `IE` and `Edge` correctly... very strange

simboonlong: ok i think SOE is not a good place to debug, the simple check from stackoverflow also printed out `chrome`

simboonlong:
needs to be verified on IE or Edge (not on SOE) if solution's working as expected

simboonlong:
stale, closing

PR #636, opened 2021-08-24
fix: circle ci context

cavacado:
This pr is to update the npm token that is no longer shared between teams.


Nebulis:
Hum it's needed here?

cavacado:
this doesnt have semantic release?
ah sorry was mistaken

PR #667, opened 2022-03-07
chore: show trusted issuer name alongside issuer domain

HJunyuan:
**What does this PR do?**
- Allow trusted issuer to have their name shown alongside issuer domain
- Trusted TLDs: `*.gov.sg` and `*.edu.sg`
- Example:
  ```text
  Certificate issued by
  NTU.EDU.SG
  NANYANG TECHNOLOGICAL UNIVERSITY
  ```

| Before | After |
|---|---|
| Screenshot 2022-03-07 at 2 54 06 PM | Screenshot 2022-03-07 at 2 54 15 PM |

PR #669, opened 2022-03-23
chore: show domain for documents issued by issuers in the registry

HJunyuan:
**What does this PR do?**
- To have consistent behaviour as #667
- I.e. show issuer domain and name for documents issued by **trusted TLDs** and **issuers in the registry**


simboonlong (approved):
lgtm

PR #701, opened 2022-11-03
chore: update from ua to ga4

zixiang2018:
## Context
To migrate Google Universal Analytics to Google Analytics 4 as "`Universal Analytics (UA) will be deprecated on July 1, 2023, which means it will stop processing data. Analytics 360 properties will stop working on October 1, 2023`"

## What this PR does
- replaced `next-ga` with `react-ga4
`- replaced using of `window.ga` with `react-ga4` functions
- removed `label` key from `Event` - GA4 does not support long string and label contains values which are already logged in our GA events
- removed sending of parameters with values - "not set", to GA events 
- updated testcases for analytics test
- replaced `GA_ID` with env variable `GTAG_ID`
- added `GTAG_ID` into dev and prod env variable 
- wrap `documentStore` with quotes before sending event to prevent GA4 from converting hexadecimal to decimal  ([related to this](https://stackoverflow.com/questions/72561923/why-is-the-variable-not-being-passed-from-the-data-layer-to-google-analytics-in))

## On analytics.google.com
  - created both dev and prod accounts 
  - added data stream for both accounts
  - added custom dimensions


src/components/Analytics/analytics.test.ts
```diff
@@ -107,58 +106,46 @@ describe("validateEvent", () => {
...
-      category: "CATEGORY",
-      action: "ACTION",
-      value: undefined,
-    });
-    expect(true).toBe(true);
+    expect(() =>
+      validateEvent({
+        category: "CATEGORY",
+        action: "ACTION",
+        value: undefined,
```
HJunyuan, line 113: Any implications if we send `undefined` as an event parameter?

src/components/Analytics/index.ts
```diff
@@ -1,16 +1,29 @@
...
+  documentStore?: string;
+  documentId?: string;
+  documentName?: string;
+  issuedOn?: string;
+  issuerName?: string;
+  registryId?: string | null;
+  rendererUrl?: string;
+  templateName?: string;
+  errors?: string;
+}
```
HJunyuan, line 20: Not too sure if it's good practice to make them all optional?

Perhaps better practice for each event to have its own interface so we know which parameters are required

src/components/DecentralisedTemplateRenderer/DecentralisedRenderer.tsx
```diff
@@ -103,18 +103,22 @@ const DecentralisedRenderer: React.FunctionComponent<DecentralisedRendererProps>
...
       const certificateData = getData(rawDocument);
-      analyticsEvent(window, {
+      analyticsEvent({
         category: "CERTIFICATE_VIEWED",
         action: certificateData.issuers.map((issuer) => issuer.id).join(","),
-        label: certificateData?.id ?? undefined,
+        options: {
+          documentId: certificateData?.id ?? undefined,
+        },
       });
```
HJunyuan, line 112: Just curious why the helper method `sendV2EventCertificateViewedDetailed()` wasn't used here?

zixiang2018: For this part, the `action` value is structured differently from `sendV2EventCertificateViewedDetailed()`. Think this is checking `isSignedWrappedV2Document`; maybe to differentiate between DID certs? Only DID certs are signed and wrapped right?

HJunyuan: DID certs are wrapped and then signed while Ethereum-issued certs are only wrapped.

Oh yes, the action in `sendV2EventCertificateViewedDetailed()` is using issuer name VS issuer id here. It's probably due to registry vs non-registry certs (i.e. OC vs OA v2)

src/components/Analytics/analytics.test.ts
zixiang2018: Just checked the behavior - as the "value" key only accepts `int`, when I sent an int value, e.g.`1`, to the event, it is recorded as `1` in the parameters. 

But if its `undefined` then the parameter, "value",  doesnt show any `undefined` values in the GA4 reports. 

In short, `undefined` seems to be "ignored" for the parameter "value"

src/components/Analytics/index.ts
zixiang2018: I was thinking whether we can make the keys here to mandatory but in the Event, we use `Partial<T>`. 

Cause for each event to have its own interface might mean more duplicated codes also

HJunyuan: Alright I suppose we can leave it like this for now

HJunyuan:
Before we can merge this to production, we'll need to confirm whether the same report can be generated in Data Studio with the data source now coming from GA4. Do you need help with that?

PR #730, opened 2023-05-22
feat: add polygon support

yapyuyou:
## What this PR does

- Add alternate network provider for non-ethereum networks
- Automatically use `matic` | `maticmum` providers based on `document.data.network` field
- Add demo cert for polygon mumbai testnet (maticmum)


src/sagas/certificate.ts
```diff
@@ -40,20 +40,45 @@ import {
...
+  );
+
+const getAlternateNetwork = (certificate: WrappedOrSignedOpenCertsDocument) => {
+  const data = opencertsGetData(certificate);
+  if (!data.network) return undefined;
+  switch (data.network?.chainId) {
+    case "137":
+      return "matic";
+    case "80001":
+      return "maticmum";
```
HJunyuan, line 71: We might need to be careful with this as it will allow testnet-issued documents to be verifiable on the production verifier (i.e. `opencerts.io`)

Perhaps we can generate a list of "allowed networks" when `IS_MAINET` is true/false under this file: https://github.com/OpenCerts/opencerts-website/blob/b42b8e2cbdaf1d48858d6215ccf7a28023fcf09c/src/config/index.ts#L13-L14

HJunyuan (changes requested):

HJunyuan (approved):

HJunyuan (approved):
🚀

PR #758, opened 2024-01-17
chore: add collaborator

zixiang2018:
## What this PR does
- Add Filpass into list of collaborators
- [Set `TRUSTED_TLDS` as `undefined` in one of the test](https://github.com/OpenCerts/opencerts-website/pull/758/files#diff-cdbd2b6c5ed1bc7034b81bf7495ff92dd947f7d3d3727ca47b1d263d19b0fb5cR124) to fix one test case which fails on our Gitlab pipeline. 
    - Test fails because in CircleCI (Github), there are no environment variables set, hence`TRUSTED_TLDS` should be `undefined` [on this line](https://github.com/OpenCerts/opencerts-website/blob/f97efd23229132b5e9ff34df6687ae6d00537112/src/components/CertificateVerifyBlock/CertificateVerifyBlock.tsx#L68).
    - But in Gitlab, `TRUSTED_TLDS` has an actual whitelist string, which was being used in the test and breaking the test.


HJunyuan (dismissed):
🚀

HJunyuan (approved):

PR #769, opened 2024-04-15
Uxd/logo fixes

witxhhaven:
- trim smu logo
- extend logo width
- add logo logic


zixiang2018 (approved):

zixiang2018:
There is a failing integration test case due to the Mumbai deprecation, I think we can merge this in first, as the migration to Amoy Testnet will need some time to develop on the OA side.

PR #785, opened 2024-07-12
feat: implement failover provider with shuffling

HJunyuan:
## Context
- Our ETH calls to Infura/Alchemy has been heavily throttled (HTTP `429 Too Many Requests`)
- `ethers.providers.FallbackProvider()` is not relevant as it does not provide network failover mechanism but to gather consensus from multiple providers

## What does this PR do?
- Temporarily install `did-resolver` and `ethr-did-resolver` to support custom failover provider class (In future, failover provider should be provided by `oa-verify` library if proved useful)
- Implement `OAFailoverProvider` class to attempt a given set of RPC endpoints to attempt
- Shuffle given set of RPC endpoints by default in order to spread the load across multiple providers
- Update default API keys from Ethers
- Maintain responsible network switching between mainnet (production) and testnet (non-production):
  - In production: Only ETH mainnet and Polygon matic are allowed
  - In non-production: Only ETH sepolia and Polygon amoy are allowed
- Resolver will always query with a mainnet provider:
  - Regardless of mainnet or testnet, OA only uses mainnet DIDs:
    - ✅ `did:ethr:0x1245e5b64d785b25057f7438f715f4aa5d965733`
    - ❌ `did:ethr:sepolia:0x1245e5b64d785b25057f7438f715f4aa5d965733`
  - As such, resolver should always resolve against a mainnet provider
  - Specifying a resolver will also prevent unnecessary "eth_chainId" calls to providers


phanshiyu (dismissed):
LGTM

phanshiyu (approved):

PR #806, opened 2024-10-29
fix: prevent exhausting failover providers on erc165 reverts

eli-lim:
### Background

1. I noticed slowness when checking my degree on the site - took about 4 - 5s to verify
2. Looking at the networks tab, I noticed it was making a bunch of requests to different public node providers in sequence

https://github.com/user-attachments/assets/dae59424-1c60-45c7-a594-743ffe0c8527

### Root cause
1. [oa-verify](https://github.com/Open-Attestation/oa-verify/blob/0492679f4f24df164d93e699a5845b30d39ab14d/src/common/utils.ts#L240) is checking if the documentStore supports some interface, compliant with the ERC165 spec. 
2. When the rpc call returns `execution reverted`, ethers will throw an error. oa-verify mitigates this with a try-catch, returning false if an error is thrown.
3. Unfortunately, website's `OAFailoverProvider` catches this deeper in the call stack (rendering the try-catch in (2) moot), and moves on to the next provider, until the list of providers is exhausted.

TLDR:
```ts
async function isBatchableDocumentStore(): boolean {
  try {
    // this is expected to throw if erc165 is not implemented, but it keeps retrying instead
    return await OAFailoverProvider.perform('call', { contract, erc165Check }) as boolean; 
  } catch (e) {
    return false;
  } 
}
```

### Fix
1. (This PR) Stop the `OAFailoverProvider` from overzealously retrying on `erc165` reverts. It can also be made more generic, to stop retrying on any call exception.
2. (Optionally) switch to use [public node providers with lower latency](https://chainlist.org/?search=ethereum)
    > Ignoring other concerns like privacy / agreements with the existing node providers.

Doing these:

https://github.com/user-attachments/assets/34217fd1-f390-4c93-b3f3-d63db6775d6c

> [!NOTE]
> An alternative is to change oa-verify to use a direct rpc call, to bypass ethers's throw-on-revert

### DIFF
--- a/src/services/failover-provider.ts
+++ b/src/services/failover-provider.ts
@@ -52,6 +52,11 @@ export class OAFailoverProvider extends providers.StaticJsonRpcProvider {
       try {
         return await this.failoverProviders[i].perform(method, params);
       } catch (error) {
+        // Don't retry if it's just a revert from an ERC165 check
+        if (this.isErc165Check(method, params, error)) {
+          return false;
+        }
+
         // If last provider
         if (i === this.failoverProviders.length - 1) {
           throw error;
@@ -62,6 +67,17 @@ export class OAFailoverProvider extends providers.StaticJsonRpcProvider {
         }
       }
     }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private isErc165Check(method: string, params: any, ethersError: any): boolean {
+    // See https://eips.ethereum.org/EIPS/eip-165
+    const erc165SupportsInterfaceSelector = "0x01ffc9a7";
+    return (
+      method === "call" &&
+      params?.transaction?.data?.startsWith(erc165SupportsInterfaceSelector) &&
+      ethersError.code === "CALL_EXCEPTION"
+    );
   }
 }
 
--- a/src/services/failover-provider.ts
+++ b/src/services/failover-provider.ts
@@ -51,9 +51,11 @@ export class OAFailoverProvider extends providers.StaticJsonRpcProvider {
     for (let i = 0; i < this.failoverProviders.length; i++) {
       try {
         return await this.failoverProviders[i].perform(method, params);
-      } catch (error) {
-        // Don't retry if it's just a revert from an ERC165 check
-        if (this.isErc165Check(method, params, error)) {
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (error: any) {
+        // Don't retry if it's a call exception. We only retry on network errors.
+        if (error?.code === "CALL_EXCEPTION") {
           return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/services/failover-provider.ts
```diff
@@ -63,6 +68,17 @@ export class OAFailoverProvider extends providers.StaticJsonRpcProvider {
...
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private isErc165Check(method: string, params: any, ethersError: any): boolean {
+    // See https://eips.ethereum.org/EIPS/eip-165
+    const erc165SupportsInterfaceSelector = "0x01ffc9a7";
+    return (
+      method === "call" &&
+      params?.transaction?.data?.startsWith(erc165SupportsInterfaceSelector) &&
+      ethersError.code === "CALL_EXCEPTION"
```
HJunyuan, line 79: Propose regex check instead for cases where `params.transaction.data` is undefined:

```typescript
(undefined).startsWith("0x01ffc9a7") // ❌ TypeError is thrown instead which may impede actual Ethers error from being thrown
/^0x01ffc9a7.*$/.test(undefined)     // ✅ false
```

```suggestion
    // Avoid retries if the function selector inside calldata is "supportsInterface(bytes4)"
    // See https://eips.ethereum.org/EIPS/eip-165
    // isBatchableDocumentStore: https://github.com/Open-Attestation/oa-verify/blob/0492679f4f24df164d93e699a5845b30d39ab14d/src/common/utils.ts#L237-L244
    // Keccak256("supportsInterface(bytes4)") = 0x01ffc9a7
    const erc165SupportsInterfaceRegex = /^0x01ffc9a7.*$/
    return (
      method === "call" &&
      erc165SupportsInterfaceRegex.test(params?.transaction?.data) &&
      ethersError?.code === "CALL_EXCEPTION"
```

src/services/failover-provider.ts
```diff
@@ -52,6 +52,11 @@ export class OAFailoverProvider extends providers.StaticJsonRpcProvider {
       try {
         return await this.failoverProviders[i].perform(method, params);
       } catch (error) {
+        // Don't retry if it's just a revert from an ERC165 check
+        if (this.isErc165Check(method, params, error)) {
+          return false;
```
HJunyuan, line 57:
```suggestion
          throw error;
```

HJunyuan (changes requested):
@eli-lim Thank you for proposing these improvements!

I've made a few minor suggestions before we proceed to merge this PR in.

src/services/failover-provider.ts
eli-lim: I've made this more generic. It shouldn't retry on any call exception, not just those originating from ERC165 reverts

HJunyuan:
@eli-lim My apologies, we would still need to perform `isErc165Check()` because the Ethers library also throws a `CALL_EXCEPTION` for other failures (e.g. when Infura is unreachable)

CALL_EXCEPTION

To avoid any further inconvenience to you, we can also take over from here to implement the changes in a separate PR.

